PR #20, opened 2025-11-06
Python serialization poc

jakub-graczyk:
## Pre-review checklist

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -170,6 +170,11 @@ impl PyRowSerializationContext {
     }
 }
 
+// Create python dict
+// kind = ColumnType
+// ...
+// if kind: ColumnType != Native
+// element_type = dict - recursively calll this method
 fn extract_type_info(py: Python, column_type: &ColumnType) -> PyResult<PyObject> {
     let type_dict = pyo3::types::PyDict::new(py);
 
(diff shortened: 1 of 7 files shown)


python/scylla/_rust/__init__.pyi
```diff
@@ -0,0 +1 @@
+__all__ = []
```
Lorak-mmk: I'm curious, why is that necessary?

src/session.rs
```diff
@@ -72,10 +128,168 @@ impl RequestResult {
...
+// ...
+// if kind: ColumnType != Native
+// element_type = dict - recursively calll this method
+fn extract_type_info(py: Python, column_type: &ColumnType) -> PyResult<PyObject> {
+    let type_dict = pyo3::types::PyDict::new(py);
+
+    match column_type {
+        // POC: Handle Native types with full structure
+        ColumnType::Native(native_type) => {
+            type_dict.set_item("kind", "native")?;
```
wprzytula: 💭 Although this might seem like redundant overhead (compared to solutions where we serialize in Rust), let's keep in mind that we will probably want to expose prepared metadata (and result metadata too, which is analogous for deserialization case) to Python user anyway.

So such API will be needed no matter which serialization flavour we go for. One important thing: if we decide to serialize in Rust, we will be able to make this type info extraction lazy, i.e., executed only on user explicit demand.

PR #21, opened 2025-11-11
Direct translation of Rust serialization API into Python

patrycja-ziemkiewicz:
## Pre-review checklist


python/scylla/_rust/column_type.pyi
```diff
@@ -0,0 +1,35 @@
...
+    column_type: Any
+
+class PyTuple:
+    element_types: List[Any]
+
+class PyUserDefinedType:
+    name: str
+    frozen: bool
+    keyspace: str
+    field_types: List[Tuple[str, Any]]
```
wprzytula: :wrench: I believe we should have an umbrella class for all CQL types, and use it instead of `Any` in all these places. Strong typing helps!

python/scylla/_rust/writers.pyi
```diff
@@ -0,0 +1,36 @@
...
+
+class PyRowSerializationContext:
+    @staticmethod
+    def from_prepared(prepared_statement: PyPreparedStatement) -> PyRowSerializationContext: ...
+    def column_count(self) -> int: ...
+    def get_columns(self) -> list[Any]: ...
+class SerializationBuffer:
+    def __init__(self) -> None: ...
+    def len(self) -> int: ...
+    def set_element_count(self, count: int) -> None: ...
```
wprzytula: :recycle: missing empty line between classes.

src/writers.rs
```diff
@@ -0,0 +1,381 @@
...
+            CellWriter::new(&mut buff)
+        } else {
+            CellWriter::new_without_size(&mut buff)
+        };
+        mut_writer.set_unset();
+    }
+
+    pub fn into_value_builder(&self) -> PyCellValueBuilder {
+        PyCellValueBuilder::new(self.buff.clone(), self.write_size)
+    }
```
wprzytula: :wrench: `into` is not appropriate for `borrowed -> owned` conversions. See [the naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

python/scylla/serializer.py
```diff
@@ -0,0 +1,342 @@
...
+        # Write element count
+        element_count_bytes = struct.pack(">i", len(value))
+        builder.append_bytes(element_count_bytes)
+
+        # Write each key-value pair
+        for k, v in value.items():
+            self.key_serializer.serialize(k, builder.make_sub_writer())
+            self.value_serializer.serialize(v, builder.make_sub_writer())
+
+        builder.finish()
```
wprzytula: `result` seems to not be used anywhere.

src/column_type.rs
```diff
@@ -0,0 +1,89 @@
+use pyo3::prelude::{PyModule, PyModuleMethods};
+use pyo3::{Bound, Py, PyAny, PyResult, Python, pyclass, pymodule};
+
+
+#[pyclass(subclass)]
+pub struct PyNativeType;
+
+#[pyclass(extends=PyNativeType)]
+pub struct Double;
```
wprzytula: :question: Is `subclass` transitive for all subclasses or not? Will `Double` be inheritable in Python or not? Which is the desired behaviour?

python/scylla/_rust/writers.pyi
```diff
@@ -0,0 +1,36 @@
+from typing import Any
+
+from .statements import PyPreparedStatement
+
+
+class PyCellOverFlowError(Exception): ...
+
+class PyRowSerializationContext:
```
wprzytula: Let's not use this superfluous `Py` prefix.

src/lib.rs
```diff
@@ -2,10 +2,12 @@ use std::sync::LazyLock;
 
 use pyo3::prelude::*;
 use tokio::runtime::Runtime;
-
+mod column_type;
```
wprzytula: This should belong to the previous commit.

src/writers.rs
```diff
@@ -0,0 +1,381 @@
...
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.buff.lock().unwrap().len()
+    }
+
+    pub fn set_element_count(&mut self, element_count: u16) {
+        self.element_count = element_count;
+    }
```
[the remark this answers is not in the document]
Lorak-mmk: It needs a runtime borrowing checks: https://pyo3.rs/v0.27.1/class.html?highlight=%26mut#bound-and-interior-mutability
Next section describes how to opt-out of this if the class doesn't need &mut: https://pyo3.rs/v0.27.1/class.html?highlight=%26mut#frozen-classes-opting-out-of-interior-mutability

I thought I read the recommendation about performing the synchronization internally, instead of using &mut and relying on PyO3 for this, but I can't find it now, so perhaps I just imagined it.

src/column_type.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub frozen: bool,
+}
+
+#[pyclass(extends=PyCollectionType)]
+pub struct Map {
+    #[pyo3(get)]
+    pub key_type: Py<PyAny>,
+    #[pyo3(get)]
+    pub value_type: Py<PyAny>,
+}
```
Lorak-mmk: Please try to define a common base class like ColumnType that could be used here instead of PyAny.

src/writers.rs
```diff
@@ -0,0 +1,381 @@
...
+                let typ = extract_type(py, col_spec.typ())?;
+
+                columns.push(typ);
+            }
+
+            Ok(columns)
+        })
+    }
+}
+
```
Lorak-mmk: This all should be precomputed on construction, not computed on each call to `get_columns`.

src/writers.rs
```diff
@@ -0,0 +1,381 @@
...
+fn extract_type(py: Python<'_>, column_type: &ColumnType) -> PyResult<Py<PyAny>> {
+    match column_type {
+        ColumnType::Native(native_type) => match native_type {
+            NativeType::Int => Py::new(
```
Lorak-mmk: Why does this return a Result?

src/writers.rs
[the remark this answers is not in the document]
Lorak-mmk: What I don't get is why manual synchronization is used for buffer, but &mut for element count.
I would expect consistency here. Either use &mut and let PyO3 synchronize, or synchronize all manually.

Also, do we need mutex? The guide mentions RefCell and it matches what I know about Python execution (objects are autmatically synchronized, so no need for Sync).

python/scylla/serializer.py
```diff
@@ -0,0 +1,342 @@
...
+from typing_extensions import override
+
+from ._rust.statements import *
+from ._rust.writers import *
+from ._rust.column_type import *
+
+INT32_MIN = -2_147_483_648
+INT32_MAX =  2_147_483_647
+INT64_MIN = -9_223_372_036_854_775_808
+INT64_MAX =  9_223_372_036_854_775_807
```
Lorak-mmk: I refuse to believe that those constants are not available in Python's std somewhere.

patrycja-ziemkiewicz:
> @patrycja-ziemkiewicz IS there a reason to keep this open?

I think no. In future We would probably use some logic from here to implement interface for python users to write their own custom serialization but I think It should be probably done in different PR.

PR #22, opened 2025-11-15
Implement Rust-side serialization

paulinaczajkowska:
## Python Value Serialization

Implemented serialization of Python values passed as `PyAny`. The approach works as follows:

- The `ColumnType` is checked to determine the expected type.
- Python values are extracted from `PyAny` and converted to Rust types.
- For collection types (e.g., `PyList`, `PyDict`, `PySet`), dedicated wrapper structs implement `SerializeValue`.
- For `NativeType`s, values are extracted to simple Rust types when possible.
- For more complex `NativeType`s that cannot be represented by simple Rust types, a corresponding `CqlValue` is created and serialized.

### DIFF
--- a/python/scylla/_rust/session.pyi
+++ b/python/scylla/_rust/session.pyi
@@ -1,6 +1,12 @@
+from typing import Any
+
 from .results import RequestResult
 from .statement import PreparedStatement, Statement
 
 class Session:
-    async def execute(self, request: PreparedStatement | Statement | str) -> RequestResult: ...
+    async def execute(
+        self,
+        request: PreparedStatement | Statement | str,
+        values: Any | None = None,
+    ) -> RequestResult: ...
     async def prepare(self, statement: Statement | str) -> PreparedStatement: ...
--- a/python/tests/test_serialization.py
+++ b/python/tests/test_serialization.py
@@ -1,0 +1,814 @@
+import ipaddress
+import uuid
+from dataclasses import asdict, dataclass
+from datetime import date, datetime, time, timezone
+from decimal import Decimal
+from typing import AsyncGenerator, Awaitable, Callable, List, Optional, Tuple
+
+import pytest
+import pytest_asyncio
+from dateutil.relativedelta import relativedelta
+from scylla.session import Session
+from scylla.session_builder import SessionBuilder
+
+
+async def set_up() -> Session:
+    builder = SessionBuilder(["127.0.0.2"], 9042)
+    session = await builder.connect()
+
+    await session.execute("""
+            CREATE KEYSPACE IF NOT EXISTS testks
+            WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1};
+        """)
+
+    await session.execute("USE testks")
+
+    return session
+
+
+@pytest_asyncio.fixture(scope="module")
+async def session():
+    session = await set_up()
+    yield session
+    await session.execute("DROP KEYSPACE testks")
+
+
+TableFactory = Callable[[str, str], Awaitable[str]]
+
+
+@pytest_asyncio.fixture
+async def table_factory(session: Session) -> AsyncGenerator[TableFactory, None]:
+    created_tables: list[str] = []
+
+    async def create_table(schema: str, name: str) -> str:
+        await session.execute(f"CREATE TABLE IF NOT EXISTS {name} ({schema});")
+        created_tables.append(name)
+        return name
+
+    yield create_table
+
+    for table in created_tables:
+        await session.execute(f"DROP TABLE IF EXISTS {table};")
+
+
+@dataclass
+class Address:
+    """Example UDT class for address"""
+
+    street: str
+    city: str
+    zip_code: int
+
+    def to_dict(self):
+        return {"street": self.street, "city": self.city, "zip_code": self.zip_code}
+
+
+@dataclass
+class Person:
+    """Example UDT class for person"""
+
+    name: str
+    age: int
+    address: Address
+
+    def to_dict(self):
+        return {
+            "name": self.name,
+            "age": self.age,
+            "address": self.address.to_dict() if self.address else None,
+        }
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_basic_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, name text, score double",
+        "text_table",
+    )
+    test_values = [1, "Test Name", 95.5]
+
+    await session.execute(f"INSERT INTO {table} (id, name, score) VALUES (?, ?, ?)", test_values)
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_ascii_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col ascii",
+        "ascii_table",
+    )
+
+    val = "HELLO"
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_boolean_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col boolean",
+        "boolean_table",
+    )
+
+    val = True
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_blob_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col blob",
+        "blob_table",
+    )
+
+    val = b"hello world"
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_date_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col date",
+        "date_table",
+    )
+
+    val = date(2004, 6, 16)
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_decimal_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col decimal",
+        "decimal_table",
+    )
+
+    val = Decimal("12.3E+7")
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_double_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col double",
+        "double_table",
+    )
+
+    val = 1.23456789
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_duration_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col duration",
+        "duration_table",
+    )
+
+    val = relativedelta(months=2, days=5, microseconds=36)
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_float_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col float",
+        "float_table",
+    )
+
+    val = 3.14
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_int_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col int",
+        "int_table",
+    )
+
+    val = 123
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_int_serialization_overflow(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col int",
+        "int_table_overflow",
+    )
+
+    val = 9999999999999999999999999
+
+    with pytest.raises(Exception):
+        await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_bigint_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col bigint",
+        "bigint_table",
+    )
+
+    val = 9999999999
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_bigint_serialization_overflow(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col bigint",
+        "bigint_table_overflow",
+    )
+
+    val = 99999999999999999999999999999999
+
+    with pytest.raises(Exception):
+        await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_text_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col text",
+        "text_table",
+    )
+
+    val = "Meow"
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_timestamp_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col timestamp",
+        "timestamp_table",
+    )
+
+    val = datetime(1, 1, 1, 23, 59, 59, 999000, tzinfo=timezone.utc)
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_inet_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col inet",
+        "inet_table",
+    )
+
+    ipv4 = ipaddress.ip_address("127.0.0.1")
+    ipv6 = ipaddress.ip_address("::1")
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, ipv4))
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, ipv6))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_smallint_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col smallint",
+        "smallint_table",
+    )
+
+    val = 12
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_smallint_serialization_overflow(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col smallint",
+        "smallint_table_overflow",
+    )
+
+    val = 999999999999999999999999
+
+    with pytest.raises(Exception):
+        await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+@pytest.mark.skip(reason="Counter currently does not support NetworkTopologyStrategy; enable when supported")
+async def test_counter_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col counter",
+        "counter_table",
+    )
+
+    val1 = 5
+    val2 = 3
+    id = 1
+
+    await session.execute(f"UPDATE {table} SET col = col + ? WHERE id = ?", (val1, id))
+    await session.execute(f"UPDATE {table} SET col = col + ? WHERE id = ?", (val2, id))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_tinyint_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col tinyint",
+        "tinyint_table",
+    )
+
+    val = 5
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_tinyint_serialization_overflow(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col tinyint",
+        "tinyint_table_overflow",
+    )
+
+    val = 99999999999999999999999
+
+    with pytest.raises(Exception):
+        await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_time_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col time",
+        "time_table",
+    )
+
+    val = time(1, 30, 45, 500000)
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_timeuuid_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col timeuuid",
+        "timeuuid_table",
+    )
+
+    val = uuid.uuid1()
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_uuid_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col uuid",
+        "uuid_table",
+    )
+
+    val = uuid.uuid4()
+
+    await session.execute(f"INSERT INTO {table} (id, col) VALUES (?, ?)", (1, val))
+    await session.execute(f"SELECT * from {table}")
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_varint_serialization(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, col varint",
+        "varint_table",
+    )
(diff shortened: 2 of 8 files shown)


src/serialize.rs
```diff
@@ -0,0 +1,222 @@
...
+            match val.downcast::<PyTuple>() {
+                Ok(py_tuple) => {
+                    for (col, val) in ctx.columns().iter().zip(py_tuple.iter()) {
+                        let wrapper = PyAnyWrapper(val.into());
+                        let sub_writer = writer.make_cell_writer();
+                        SerializeValue::serialize(&wrapper, col.typ(), sub_writer)?;
+                    }
+                    Ok(())
+                }
+                Err(_) => Err(invalid_input("expected Python tuple")),
```
wprzytula: 🤔 We're returning `SerializationError`, whereas this looks more like a Python `TypeError` to me (because the tuple is expected as the container for bound values).

src/serialize.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+impl<'a> PyListWrapper<'a> {
+    fn new(value: &PyAnyWrapper, list_type: &'a ColumnType<'a>) -> PyResult<Self> {
+        Python::with_gil(|py| {
+            let val = value.bind(py);
+            let list: &Bound<PyList> = val.downcast::<PyList>()?;
+            Ok(PyListWrapper {
+                inner: list.clone().unbind(),
```
wprzytula: ❓ What does `clone()` on `Bound` do?

src/session.rs
```diff
@@ -29,7 +29,28 @@ impl Session {
...
+            .spawn(async move {
+                session_clone
+                    .query_unpaged(request_string, PyAnyWrapper(values))
+                    .await
+                    .map_err(|e| {
+                        PyRuntimeError::new_err(format!("Failed to deserialize metadata: {}", e))
+                    })
+            })
+            .await
+            .expect("Driver should not panic")?;
```
wprzytula: 🔧 It makes little sense to spawn a task on a runtime only to await it straightaway. I'd suggest awaiting the query future directly, without using `RUNTIME`.

Makefile
```diff
@@ -0,0 +1,105 @@
+COMPOSE := docker compose -f docker-compose.yml
+
+.PHONY: all
+all: test
+
+.PHONY: static
+static: fmt-check check check-without-features check-all-features clippy clippy-all-features clippy-cpp-rust
```
wprzytula: 🔧 As we requested, please extract makefile and CI-related changes to a separate PR and then base your PR on top of it.

wprzytula:
In general, this approach looks promising. Little Python, clear Rust. There are two caveats though:
1. This approach does not support custom serialization for user types, so it would have to be done using a separate path.
2. We definitely need to benchmark this approach, because heavy use of PyO3 may introduce large overhead.

src/serialize.rs
```diff
@@ -0,0 +1,222 @@
...
+use scylla_cql::_macro_internal::{RowWriter, SerializeRow, SerializeValue, WrittenCellProof};
+use scylla_cql::frame::response::result::{CollectionType, UserDefinedType};
+use scylla_cql::serialize::CellWriter;
+use scylla_cql::serialize::SerializationError;
+use std::ops::Deref;
+use std::sync::Arc;
+
+fn invalid_input(msg: &str) -> SerializationError {
+    SerializationError::new(std::io::Error::new(std::io::ErrorKind::InvalidInput, msg))
+}
```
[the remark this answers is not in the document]
wprzytula: You may find copying [these helper functions](https://github.com/scylladb/scylla-rust-driver/blob/e0f152c012bccbe0c975fb18289ced2af20836c7/scylla-cql/src/serialize/value.rs#L1087C4-L1087C8) from Rust Driver to the wrapper helpful.

src/serialize.rs
```diff
@@ -0,0 +1,222 @@
...
+        });
+
+        SerializeValue::serialize(&items, self.list_type, writer)
+    }
+}
+
+pub struct PyUdtWrapper<'a> {
+    pub inner: Py<PyDict>,
+    pub definition: &'a Arc<UserDefinedType<'a>>,
+}
```
wprzytula: ditto: `pub`

Lorak-mmk:
I see that PR description and checklist are not filled. Please fix that.

src/serialize.rs
```diff
@@ -0,0 +1,391 @@
...
+
+            Err(SerializationError::new(PyTypeError::new_err(
+                "expected Python tuple or list",
+            )))
+        })
+    }
+
+    fn is_empty(&self) -> bool {
+        Python::with_gil(|py| self.is_none(py))
+    }
```
Lorak-mmk: Can you also store `Bound` in `PyAnyWrapperRow`, as you do in `PyAnyWrapper`? That would allow you to avoid locking and unlocking gil multiple times.

src/serialize.rs
```diff
@@ -0,0 +1,391 @@
...
+                CollectionType::List(element_type) => {
+                    let list = match PyListWrapper::new(self, typ) {
+                        Ok(list) => list,
+                        Err(_) => {
+                            let expected = self.map_type_to_expected()?;
+                            return Err(SerializationError::new(mk_typck_err::<PyList>(
+                                typ,
+                                BuiltinTypeCheckErrorKind::MismatchedType { expected },
+                            )));
+                        }
```
Lorak-mmk: One downside od this approach (map_type_to_expected) is that you have one place where you check constraints in order to serialize, and other where you generate error messages. There is no mechanism to prevent those two places getting out of sync. Example: if you pass Python list to a `TEXT` column, you will go into the `unreachable!()` branch in `map_type_to_expected` and panic.

I'm not sure what is the best way to solve that. One idea that comes to my mind is to reverse the logic of serialization a bit. Currently you are matching on column type, and then trying to cast-down the Python type.
The alternative way is to match on supported Python types, and then check if the column type is right for the type. With this approach, we immediately know what the Python type is, so we know what are the correct CQL types for error message.
I'm not sure if it is actually better. It would require more downcasting (for which we should then use `cast` instead of `extract`: https://pyo3.rs/v0.27.2/performance.html#extract-versus-cast ), but its hard to guess what the performance impact could be.

Maybe there is a different way to avoid the issues with `map_type_to_expected`? @wprzytula

src/serialize.rs
```diff
@@ -0,0 +1,391 @@
...
+                        },
+                    )));
+                }
+
+                for (col, val) in ctx.columns().iter().zip(py_list.iter()) {
+                    let wrapper = PyAnyWrapper(val);
+                    let sub_writer = row_writer.make_cell_writer();
+                    SerializeValue::serialize(&wrapper, col.typ(), sub_writer)?;
+                }
+                return Ok(());
```
Lorak-mmk: There is quite a lot of duplication between the two cases. Is there no way to avoid that?

src/serialize.rs
```diff
@@ -0,0 +1,391 @@
...
+        match native_type {
+            NativeType::Boolean => self.serialize_native::<bool>(typ, cell_writer),
+            NativeType::BigInt => self.serialize_native::<i64>(typ, cell_writer),
+            NativeType::Double => self.serialize_native::<f64>(typ, cell_writer),
+            NativeType::Int => self.serialize_native::<i32>(typ, cell_writer),
+            NativeType::Text => self.serialize_text(typ, cell_writer),
+            _ => unimplemented!("other native types not supported yet"),
+        }
+    }
+
```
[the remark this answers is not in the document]
wprzytula: Agreed.

src/serialize.rs
wprzytula: My first thought is that matching on ColumnType is more orderly (and intuitively should be more performant as well) than a long list of downcast attempts on Python object. Therefore, I'd keep this matching approach.

Cargo.toml
```diff
@@ -9,5 +9,10 @@ crate-type = ["cdylib"]
...
 [dependencies]
 pyo3 = { version = "0.25.1", features = ["extension-module", "experimental-async"] }
-scylla = "1.3.1"
+#scylla = "1.3.1"
+scylla = { git = "https://github.com/paulinaczajkowska/scylla-rust-driver", branch = "main" }
 tokio = { version = "1.42.0", features = ["full"] }
+thiserror = "2.0.6"
+uuid = { version = "1.11.0", features = ["v4"] }
+#scylla-cql = "1.3.1"
+scylla-cql = { git = "https://github.com/paulinaczajkowska/scylla-rust-driver", branch = "main" }
```
wprzytula: @Lorak-mmk We shouldn't encourage using private forks of Rust Driver for such purposes. IMO, we should either:
1. Have a fork of Rust Driver in this org and update it with relevant APIs when needed, and sync it with upstream / merge to upstream from time to time, or
2. Merge relevant APIs to upstream immediately.

python/scylla/_rust/session.pyi
```diff
@@ -1,7 +1,8 @@
 from .statement import PreparedStatement
+from typing import Any, Optional
 
 class Session:
-    async def execute(self, request: PreparedStatement | str) -> RequestResult: ...
+    async def execute(self, request: PreparedStatement | str, values: Optional[Any] = None) -> RequestResult: ...
```
wprzytula: 🔧 I could see that we accept both `PreparedStatement` and unprepared statement.

src/serialize/row.rs
```diff
@@ -0,0 +1,117 @@
...
+            Err(SerializationError::new(mk_typck_err_named_row(
+                name,
+                scylla::serialize::row::BuiltinTypeCheckErrorKind::WrongColumnCount {
+                    rust_cols: row_len,
+                    cql_cols: cols_len,
+                },
+            )))
+        } else {
+            Ok(())
+        }
```
wprzytula: :pick: The main flow (without errors) should be linear. On error we should prefer early return.

```suggestion
        if row_len != cols_len {
            return Err(SerializationError::new(mk_typck_err_named_row(
                name,
                scylla::serialize::row::BuiltinTypeCheckErrorKind::WrongColumnCount {
                    rust_cols: row_len,
                    cql_cols: cols_len,
                },
            )));
        }
        
        Ok(())
```

src/serialize/row.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    fn serialize_element(
+        col: &ColumnSpec,
+        val: Bound<PyAny>,
+        row_writer: &mut RowWriter,
+    ) -> Result<(), SerializationError> {
+        let wrapper = PyAnyWrapper(val);
```
wprzytula: ❓ Does `PyAnyWrapper` need ownership over `Bound<PyAny>`? Perhaps a reference would be enough?
Let's not make methods take non-`Copy` data by move if borrowing suffice.

src/serialize/row.rs
```diff
@@ -0,0 +1,117 @@
...
+                Err(SerializationError::new(PyTypeError::new_err(
+                    "expected Python tuple, list or dict",
+                )))
+            }
+        })
+    }
+
+    fn is_empty(&self) -> bool {
+        Python::with_gil(|py| self.is_none(py))
+    }
```
wprzytula: 💭 Hmm, `is_none` seems to only check Python `is None`. Here, we would prefer to treat the following as empty rows, too:
- `{}`,
- `[]`,
- `tuple()`.

src/serialize/value.rs
```diff
@@ -0,0 +1,906 @@
+use pyo3::prelude::*;
+use pyo3::types::{PyBytes, PyDict, PyList, PyMapping, PySet, PyTuple};
+use pyo3::{Bound, PyAny, PyErr, PyResult};
+use scylla_cql::_macro_internal::{
+    CellWriter, ColumnType, SerializationError, SerializeValue, UdtTypeCheckErrorKind,
+    WrittenCellProof,
+};
```
wprzytula: ditto

src/serialize/value.rs
```diff
@@ -0,0 +1,906 @@
...
+    PythonError(#[from] PyErr),
+
+    #[error("Unknown native type")]
+    UnknownNativeType,
+
+    #[error("Unknown collection type")]
+    UnknownCollectionType,
+
+    #[error("Unknown collection type")]
+    UnknownColumnType,
```
wprzytula: 🔧 Copy-paster error: `Unknown collection type` in both places.

src/serialize/value.rs
```diff
@@ -0,0 +1,906 @@
...
+    fn mismatched_type_error(&self, name: &'static str, typ: &ColumnType) -> SerializationError {
+        let expected = match self.map_type_to_expected() {
+            Ok(expected) => expected,
+            Err(e) => return e,
+        };
+        SerializationError::new(mk_typck_err_named(
+            name,
+            typ,
+            BuiltinTypeCheckErrorKind::MismatchedType { expected },
+        ))
```
wprzytula: 😮 So here we can error out when trying to construct an error. Very Python things...

src/serialize/value.rs
```diff
@@ -0,0 +1,906 @@
...
+
+                let date = CqlDate(value);
+                date.serialize(typ, cell_writer)
+            }
+            NativeType::Decimal => {
+                let value = self.extract::<(Vec<u8>, i32)>().map_err(|_| {
+                    self.mismatched_type_error(std::any::type_name::<CqlDecimal>(), typ)
+                })?;
+
+                let decimal = CqlDecimal::from_signed_be_bytes_and_exponent(value.0, value.1);
```
wprzytula: ♻️ Prefer deconstruction over tuple indexed access: `let (bytes, exponent) = ...`

src/serialize/value.rs
```diff
@@ -0,0 +1,906 @@
...
+                    months: value.0,
+                    days: value.1,
+                    nanoseconds: value.2,
+                };
+                duration.serialize(typ, cell_writer)
+            }
+            NativeType::Float => self.serialize_native::<f32>(typ, cell_writer),
+            NativeType::Int => self.serialize_native::<i32>(typ, cell_writer),
+            NativeType::BigInt => self.serialize_native::<i64>(typ, cell_writer),
+            NativeType::Text => self.serialize_text(typ, cell_writer),
```
wprzytula: ♻️ Can you please group `NativeType` by categories? I.e., int types ordered by ascending size, then floats ordered by ascending size, then boolean, then...

src/serialize/value.rs
```diff
@@ -0,0 +1,906 @@
...
+                    self.mismatched_type_error(std::any::type_name::<CqlTimeuuid>(), typ)
+                })?;
+
+                let value = value.downcast::<PyBytes>().map_err(|_| {
+                    self.mismatched_type_error(std::any::type_name::<CqlTimeuuid>(), typ)
+                })?;
+
+                let bytes = value.extract::<[u8; 16]>().map_err(|_| {
+                    self.mismatched_type_error(std::any::type_name::<CqlTimeuuid>(), typ)
+                })?;
```
wprzytula: 💭 We had `packed` attribute in Inet and here we have `bytes`. Where do you take them from?

wprzytula (changes requested):
I haven't managed to go through the whole code, as the amount of it is massive, but there are already quite many comments.

Cargo.toml
Lorak-mmk: And why do you need scylla-cql?

src/session.rs
```diff
@@ -17,18 +18,31 @@ pub(crate) struct Session {
 
 #[pymethods]
 impl Session {
-    async fn execute(&self, request: PyObject) -> PyResult<RequestResult> {
+    #[pyo3(signature = (request, values = None))]
+    async fn execute(
+        &self,
+        request: PyObject,
```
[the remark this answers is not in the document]
Lorak-mmk: `PyObject` is just `Py<PyAny>` I think

src/serialize/value.rs
```diff
@@ -0,0 +1,906 @@
...
+
+        let columns = match python_type_name {
+            "int" => INT_COLUMNS,
+            "float" => FLOAT_COLUMNS,
+            "bool" => BOOL_COLUMNS,
+            "str" => STRING_COLUMNS,
+
+            // Only a subset of Python types is currently supported. More can be added after
+            // deciding on error handling approach.
+            _ => unimplemented!("{:?}", python_type_name),
```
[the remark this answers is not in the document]
Lorak-mmk: Yes, I mentioned this in previous review

Cargo.toml
Lorak-mmk: I agree with that fully. I invite you to propose a design for scylla-cql that allows this, we can then implement it in 2.0.

src/session.rs
```diff
@@ -1,36 +1,90 @@
...
+
 #[pymethods]
 impl Session {
-    async fn execute(&self, request: PyObject) -> PyResult<RequestResult> {
+    #[pyo3(signature = (request, values = None))]
+    async fn execute(
+        &self,
+        request: PyObject,
+        values: Option<Py<PyAny>>,
+    ) -> PyResult<RequestResult> {
```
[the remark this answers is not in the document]
wprzytula: `PyObject` is deprecated in the newest PyO3. Let's avoid it.

src/session.rs
```diff
@@ -1,36 +1,90 @@
...
+        let row_kind = try_into_row_kind(values)?;
+
         if let Ok(prepared) = Python::with_gil(|py| {
             let scylla_prepared = request.extract::<Py<PreparedStatement>>(py)?;
             Ok::<Py<PreparedStatement>, PyErr>(scylla_prepared)
         }) {
             let result = self
                 .session_spawn_on_runtime(async move |s| {
```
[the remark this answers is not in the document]
wprzytula: So the whole body of `if let Ok(prepared)` and others alike would go into `execute_prepared()`.

src/serialize/row.rs
```diff
@@ -0,0 +1,187 @@
...
+        row_writer: &mut RowWriter,
+    ) -> Result<(), SerializationError> {
+        Python::with_gil(|py| {
+            let row = self.bind(py);
+
+            length_equality_check(
+                row.len(),
+                ctx.columns().len(),
+                std::any::type_name::<PyList>(),
+            )?;
```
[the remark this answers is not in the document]
wprzytula: Actually, `length_equality_check` would just call generic `mk_typck_err_row` function.

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+
+    #[error("Unknown native type")]
+    UnknownNativeType,
+
+    #[error("Unknown collection type")]
+    UnknownCollectionType,
+
+    #[error("Unknown column type")]
+    UnknownColumnType,
+}
```
wprzytula: :wrench: Can we include the type name in those `Unknown type` variants?

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+            NativeType::Varint => {
+                let bits = self
+                    .call_method0("bit_length")
+                    .and_then(|x| x.extract::<usize>())
+                    .map_err(|_| self.mismatched_type_error::<CqlVarint>(typ))?;
+
+                let len = max(1, (bits + 8) / 8);
+
+                let py_dict = PyDict::new(self.py());
+                py_dict.set_item("signed", true).unwrap();
```
wprzytula: :wrench: Justify the `unwrap()`. Driver should not panic in such code.

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+    VectorSerializationErrorKind, serialize_next_variable_length_elem,
+};
+use scylla_cql::serialize::writers::CellValueBuilder;
+use scylla_cql::value::{CqlVarintBorrowed, ValueOverflow};
+use std::cmp::max;
+use std::net::IpAddr;
+use std::ops::Deref;
+use std::sync::Arc;
+use thiserror::Error;
+use time_03::{OffsetDateTime, Time};
```
wprzytula: :wrench: I can see you're using `time` types. Can we use `chrono` types instead? I believe they are preferred.

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+                uuid.serialize(typ, cell_writer)
+            }
+
+            _ => Err(SerializationError::new(
+                PythonDriverSerializationError::UnknownNativeType,
+            )),
+        }
+    }
+
+    // In future PyO3 release FromPyObjectBound can be used instead of FromPyObject.
```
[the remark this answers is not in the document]
wprzytula: @jakub-graczyk If you open the PR that bumps PyO3 version soon, then this TODO will probably be resolvable yet in this PR.

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+        let value = match self.extract::<T>() {
+            Ok(val) => val,
+            Err(_) => {
+                let expected = self.map_type_to_expected()?;
+                return Err(SerializationError::new(mk_typck_err::<T>(
+                    typ,
+                    BuiltinTypeCheckErrorKind::MismatchedType { expected },
+                )));
+            }
+        };
```
wprzytula: :recycle: Please use `let-else`.

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+        value.serialize(typ, cell_writer)
+    }
+
+    fn serialize_types<'b>(
+        &self,
+        typ: &ColumnType,
+        cell_writer: CellWriter<'b>,
+    ) -> Result<WrittenCellProof<'b>, SerializationError> {
```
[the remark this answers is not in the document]
wprzytula: How would you describe the purpose of this method?

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+                                typ,
+                                BuiltinTypeCheckErrorKind::MismatchedType { expected },
+                            )));
+                        }
+                    };
+                    list.serialize(typ, cell_writer)
+                }
+
+                CollectionType::Map(_, _) => {
+                    let map = match PyMappingWrapper::new(self) {
```
wprzytula: :wrench: `PyMapWrapper` would be a more consistent name.

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+
+impl<'a, 'py> PyListWrapper<'a, 'py> {
+    fn new(value: &PyAnyWrapper<'a, 'py>) -> PyResult<Self> {
+        let list: &Bound<PyList> = value.downcast::<PyList>()?;
+        Ok(PyListWrapper {
+            inner: list,
+            // inner: Bound::clone(list),
+        })
+    }
+}
```
wprzytula: :recycle: Leftover commented out code.

src/serialize/value.rs
```diff
@@ -0,0 +1,1075 @@
...
+
+fn serialize_sequence<'t, 'b, 'py>(
+    rust_name: &'static str,
+    len: usize,
+    iter: impl Iterator<Item = Bound<'py, PyAny>>,
+    typ: &ColumnType,
+    writer: CellWriter<'b>,
+) -> Result<WrittenCellProof<'b>, SerializationError> {
```
wprzytula: :wrench: This can also be generic over `T: Any` instead of accepting `rust_name`.

src/serialize/value.rs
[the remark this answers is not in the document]
paulinaczajkowska: image
PyO3 doesn’t specify a concrete version for this change.

python/tests/test_serialization.py
```diff
@@ -0,0 +1,682 @@
...
+
+
+async def set_up() -> Session:
+    builder = SessionBuilder(["127.0.0.2"], 9042)
+    session = await builder.connect()
+
+    await session.execute("""
+            CREATE KEYSPACE IF NOT EXISTS testks
+            WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};
+        """)
```
wprzytula: :wrench: Please don't use `SimpleStrategy`.
```suggestion
    await session.execute("""
            CREATE KEYSPACE IF NOT EXISTS testks
            WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1};
        """)
```

src/serialize/value.rs
```diff
@@ -0,0 +1,1050 @@
...
+                // but two's-complement representation in big-endian order (CqlVarintBorrowed format)
+                // requires 2 bytes, since a single signed byte covers only the range -128..127.
+                let bytes_len = bits_len / 8 + 1;
+
+                // TODO: I think there might be a way to create PyDict only once
+                // and reuse it in every `Varint` serialization.
+                let py_dict = PyDict::new(self.py());
+                py_dict.set_item("signed", true).map_err(|e| {
+                    SerializationError::new(PythonDriverSerializationError::PythonError(e))
+                })?;
```
wprzytula: :thought_balloon: Good observation - this should be doable. Maybe some kind of a `LazyLock` (or `OnceLock`) would allow for this?

src/serialize/value.rs
```diff
@@ -0,0 +1,1050 @@
...
+            NativeType::Float => self.serialize_native::<f32>(typ, cell_writer),
+            NativeType::Double => self.serialize_native::<f64>(typ, cell_writer),
+            NativeType::Decimal => {
+                let value = self
+                    .extract::<BigDecimal>()
+                    .map_err(|_| self.mismatched_type_error::<CqlDecimal>(typ))?;
+
+                let decimal = CqlDecimal::try_from(value)
+                    .map_err(|_| SerializationError::new(ValueOverflow))?;
+                decimal.serialize(typ, cell_writer)
```
wprzytula: :wrench: You should be able to directly serialize `BigDecimal`, as it's supported by the Rust Driver.

src/serialize/value.rs
```diff
@@ -0,0 +1,1050 @@
...
+            }
+
+            // Datatime types.
+            NativeType::Date => {
+                let value = self
+                    .extract::<NaiveDate>()
+                    .map_err(|_| self.mismatched_type_error::<CqlDate>(typ))?;
+
+                let date = CqlDate::from(value);
+                date.serialize(typ, cell_writer)
```
wprzytula: :wrench: Ditto: `NaiveDate` is serializable directly.

src/serialize/value.rs
```diff
@@ -0,0 +1,1050 @@
...
+            NativeType::Time => {
+                let value = self
+                    .extract::<NaiveTime>()
+                    .map_err(|_| self.mismatched_type_error::<CqlTime>(typ))?;
+
+                let time: CqlTime = value
+                    .try_into()
+                    .map_err(|_| self.mismatched_type_error::<CqlTime>(typ))?;
+
+                time.serialize(typ, cell_writer)
```
wprzytula: :wrench: ditto: `NaiveTime` should be serializable directly.

src/serialize/value.rs
```diff
@@ -0,0 +1,1050 @@
...
+                        python_collection_name.extract::<&str>().map_err(|e| {
+                            SerializationError::new(PythonDriverSerializationError::PythonError(e))
+                        })?;
+
+                    Err(SerializationError::new(
+                        PythonDriverSerializationError::UnknownCollectionType(
+                            python_collection_name.to_string(),
+                        ),
+                    ))
+                }
```
wprzytula: :wrench: The same problem: you're reporting Python type name as unknown, while the problem is with the CQL CollectionType.

src/serialize/value.rs
```diff
@@ -0,0 +1,1050 @@
...
+        }
+
+        self.serialize_arbitrary_value(typ, cell_writer)
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct PyListWrapper<'a, 'py> {
+    pub(crate) inner: &'a Bound<'py, PyList>,
+}
```
wprzytula: :wrench: ditto about `pub(crate)`. Is it needed?

src/serialize/value.rs
```diff
@@ -0,0 +1,1050 @@
...
+            .map_err(|_| mk_ser_err::<PyMapping>(typ, BuiltinSerializationErrorKind::SizeOverflow))
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct PyListVectorWrapper<'py, 'a> {
+    pub(crate) inner: &'a Bound<'py, PyList>,
+    pub(crate) dimension: u16,
+    pub(crate) element_type: &'a ColumnType<'a>,
+}
```
wprzytula: ditto: visibility

Cargo.toml
```diff
@@ -12,10 +12,11 @@ pyo3 = { version = "0.27.2", features = ["extension-module", "experimental-async
...
 scylla-cql = { git = "https://github.com/scylladb-zpp-2025-python-rs-driver/scylla-rust-driver", features = ["chrono-04", "bigdecimal-04", "num-bigint-04"]}
-bigdecimal-04 = { package = "bigdecimal", version = "0.4" }
-chrono_04 = { package = "chrono", version = "0.4" }
-num-bigint-04 = { package = "num-bigint", version = "0.4" }
 
 yoke = "0.8.1"
 stable_deref_trait = "1.2"
-uuid = "1.18.1"
\ No newline at end of file
+thiserror = "2.0.6"
```
Lorak-mmk: Renaming dependencies should be done in a separate commit.

src/session.rs
```diff
@@ -1,35 +1,69 @@
...
 #[pyclass]
 pub(crate) struct Session {
     pub(crate) _inner: Arc<scylla::client::session::Session>,
 }
 
+fn try_into_row(values: Option<Py<PyAny>>) -> PyResult<Option<PyAnyWrapperRow>> {
+    let Some(val) = values else {
+        return Ok(None);
+    };
+
```
Lorak-mmk: Nit: Imo this would be cleaner if `values` was not an `Option`. The only reason it is, is because the only caller has an `Option`. Instead, the caller could call `map` on its option.

src/serialize/value.rs
Lorak-mmk: Nit: I'm not a fan of ordering in this file. There are two typically used orderings: external to internal, and internal to external. Internal = smaller functions. leaf functions etc. External = public functions, functions that call the more internal functions.

I don't want to discuss which is better. What I want to note is that you use neither. File has first some error-related data, then serialization of native types, then definition of PyAnyWrapper, and SerializeValue impl (both are definitely most external!), and then serialization of collection.

Please try to use more reader-friendly ordering. Right now if I want to read external to internal, I need to start in the middle, then read lower portion, then upper portion, then beginning and end (for error-related helpers). If I want to read internal to external, I need to do that in reverse. Both approaches are not convenient.

src/serialize/value.rs
```diff
@@ -0,0 +1,985 @@
...
+                    .cast::<PyBytes>()
+                    .map_err(|_| self.mismatched_type_error::<Uuid>(typ))?;
+
+                let bytes = value
+                    .extract::<[u8; 16]>()
+                    .map_err(|_| self.mismatched_type_error::<Uuid>(typ))?;
+
+                let uuid = Uuid::from_bytes(bytes);
+                uuid.serialize(typ, cell_writer)
+            }
```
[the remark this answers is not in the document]
Lorak-mmk: Also, if you don't need the error returned you should call `cast` instead of `extract`, because it is more performant in error path according to PyO3 book: https://pyo3.rs/v0.27.2/performance.html#extract-versus-cast . This is a general comment, not only for this place.

src/serialize/value.rs
```diff
@@ -0,0 +1,985 @@
...
+        | ColumnType::Collection {
+            typ: CollectionType::Set(elt),
+            ..
+        } => elt,
+        _ => {
+            return Err(mk_typck_err::<T>(
+                typ,
+                SetOrListTypeCheckErrorKind::NotSetOrList,
+            ));
+        }
```
Lorak-mmk: When it is possible to enter this error path?

Lorak-mmk (changes requested):

wprzytula (approved):

PR #23, opened 2025-11-16
Python typed serialization

jakub-graczyk:
## Pre-review checklist

### DIFF
--- a/python/tests/test_serialization.py
+++ b/python/tests/test_serialization.py
@@ -200,6 +200,15 @@ async def test_context_udt_types():
 
     # Drop and recreate to ensure clean state
     _ = await session.execute("DROP TABLE IF EXISTS ctx_udt_types")
+    _ = await session.execute("DROP TYPE IF EXISTS address")
+
+    _ = await session.execute("""
+        CREATE TYPE address (
+            street text,
+            city text,
+            zip_code int
+        )
+    """)
 
     _ = await session.execute("""
         CREATE TABLE ctx_udt_types (


src/session.rs
```diff
@@ -36,12 +44,48 @@ impl Session {
...
+            .spawn(async move {
+                session_clone.prepare(request_string).await.map_err(|e| {
+                    PyRuntimeError::new_err(format!("Failed to deserialize metadata: {}", e))
+                })
+            })
+            .await
+            .expect("Driver should not panic")?;
+        return Ok(PyPreparedStatement {
+            _inner: (Arc::new(result)),
+        });
```
wprzytula: :question: What for do we spawn the task on tokio runtime? Can't we await it specifically on the calling thread?

src/session.rs
```diff
@@ -36,12 +44,48 @@ impl Session {
...
+        };
+
+        let (result, _) = RUNTIME
+            .spawn(async move {
+                session_clone
+                    .execute(&prep, &ser, None, PagingState::start())
+                    .await
+                    .map_err(|e| PyRuntimeError::new_err(format!("Failed to execute query: {}", e)))
+            })
+            .await
```
wprzytula: ditto: it makes no sense to spawn a task on the runtime only to await it straight after.

python/scylla/session.py
```diff
@@ -1 +1,33 @@
...
+    RequestResult,
+    Session,
+)
+
+__all__ = ["Session"]
+
+
+async def execute_with_column_spec(
+    self: Session, prepared: PyPreparedStatement, values: ValueList
+) -> RequestResult:
```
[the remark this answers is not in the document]
Lorak-mmk: Let's introduce a formatter and stick to what it does.

src/session.rs
Lorak-mmk: I invite you to checkout the PR locally to try and do this. It will be a good exercise to better understand limitations of async ffi.

wprzytula: In Node.js Driver, the binding library hides task spawning, so I don't even know.
But you're right, this limitation makes perfect sense: Python threads most probably cannot be tied to Rust reactor, because that would conflict with Python reactor.

Lorak-mmk:
@jakub-graczyk Is there a reason to keep this open?

PR #24, opened 2025-11-16
Developer aid + CI

jakub-graczyk:
Added Makefile, docker-compose, ruff, basedpyright.
Makefile supports docker, rust + python type checking, formatting, static linting and tests.

Added a pre-commit hook which enforces static checks.

Fixed source to pass static checks.

Updated CI.yml to run static checks + smoke tests on every push.


README.md
```diff
@@ -8,7 +8,14 @@ This project is in early devlopment. Not ready for production usage.
...
 
-No documentation yet :(
+### Setting Up Git Hooks
+
+To ensure code quality, we recommend enabling the pre-commit hook that runs static checks and automatic fixes before each commit:
+
+```bash
+cat scripts/pre-commit.sh > .git/hooks/pre-commit
+chmod +x .git/hooks/pre-commit
+```
```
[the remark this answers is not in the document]
wprzytula: I'm just a bit afraid it may slow down development if you tend to create many temporary commits and only then rewrite/squash them (which is my approach).

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "scylla"
-requires-python = ">=3.8"
+requires-python = ">=3.12"
```
Lorak-mmk: In general we want to support all not-EOL versions of Python. What is the reason for raising the requirement to 3.12?

pyproject.toml
```diff
@@ -18,7 +18,13 @@ features = ["pyo3/extension-module"]
...
 
 [dependency-groups]
-dev = ["pytest", "pytest-asyncio","maturin"]
+dev = [
+    "pytest",
+    "pytest-asyncio",
+    "maturin",
+    "ruff>=0.14.5",
+    "basedpyright>=1.33.0",
+]
```
Lorak-mmk: Ah, the tools are added in the second commit. Could you move this change to the first commit?

Lorak-mmk (changes requested):

python/tests/test_imports.py
```diff
@@ -1,8 +1,13 @@
...
+@pytest.mark.smoke
 def test_imports():
-    from scylla.session_builder import SessionBuilder as X
-    from scylla.session import Session as X
-    import scylla as X
-    import scylla.session as X
-    import scylla.session_builder as X
-    from scylla import session as X
-    from scylla import session_builder as X
+    import scylla as X  # pyright: ignore[reportUnusedImport]
```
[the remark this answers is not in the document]
jakub-graczyk: Not that I am aware

pyproject.toml
Lorak-mmk: Unresolving. I don't see answer to my question about version restrictions. In other words, why `ruff>=0.14.5` and not just `ruff`?

pyproject.toml
```diff
@@ -29,3 +35,14 @@ cache-keys = [
...
+[tool.ruff]
+line-length = 120
+
+[tool.basedpyright]
+typeCheckingMode = "strict"
+
+[tool.pytest.ini_options]
+markers = [
+    "smoke: smoke tests that require database",
+]
```
Lorak-mmk: I see that you still define some tests as tests that require database.
As discussed on the meeting, let's have a `requires_db` marker and put in on tests that need Scylla.

python/tests/test_imports.py
```diff
@@ -1,8 +1,9 @@
+# ruff: noqa: F811 F401
```
wprzytula: :question: What is this about?
:wrench: Please always provide in-code explanations for turning off any lints.

python/tests/test_imports.py
```diff
@@ -1,8 +1,13 @@
...
-    from scylla.session_builder import SessionBuilder as X
-    from scylla.session import Session as X
-    import scylla as X
-    import scylla.session as X
-    import scylla.session_builder as X
-    from scylla import session as X
-    from scylla import session_builder as X
+    import scylla as X  # pyright: ignore[reportUnusedImport]
+    import scylla.session as X  # pyright: ignore[reportUnusedImport]
+    import scylla.session_builder as X  # pyright: ignore[reportUnusedImport]
```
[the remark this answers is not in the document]
wprzytula: :question: Why does Copilot keep complaining about this?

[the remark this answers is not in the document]
wprzytula: Isn't this complaint justified?

scripts/pre-commit.sh
```diff
@@ -0,0 +1,22 @@
...
+echo "Running static checks (make static)..."
+echo ""
+
+# Run make static
+if ! make static; then
+    echo ""
+    echo "❌ Static checks failed!"
+    echo ""
+    echo "Your commit has been blocked because static checks did not pass."
+    echo "Possible fixes have been applied but some are remaining"
```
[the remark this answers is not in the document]
wprzytula: +1

.github/workflows/CI.yml
```diff
@@ -1,181 +1,97 @@
...
+    static-checks:
+        name: Static Checks
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+
+            - name: Install Rust toolchain
+              uses: dtolnay/rust-toolchain@stable
+              with:
+                  components: rustfmt, clippy
```
Lorak-mmk: In Rust Driver we use `actions-rust-lang/setup-rust-toolchain@v1`. Let's be consistent with that.

wprzytula (approved):

PR #25, opened 2025-11-25
Deserialization

patrycja-ziemkiewicz:
Implemented  deserialization: native types such as int, str, and float are first deserialized into Rust primitives and then converted into their corresponding Python types, while heap-allocated and collection types are deserialized directly into Python objects. 
Implemented RowsResult as an iterator over rows, each of which is mapped through the user-selected row representation via the row factory. 
Added full RowFactory support, with the default factory producing Python dicts and custom factories enabling arbitrary row shapes. 
Added ColumnIterator, which yields (column_name, column_value) pairs. When creating a RowsResult, the user may provide their own class extending RowFactory and implementing a build method. This method receives the ColumnIterator and can construct any custom row shape the user wants, using the iterator to access columns.
Added initial tests covering basic functionality.

### DIFF
--- a/python/tests/test_deserialization.py
+++ b/python/tests/test_deserialization.py
@@ -1,0 +1,1185 @@
+from typing import Any, Callable, Awaitable, AsyncGenerator, Dict, Tuple, Set, List
+
+from scylla._rust.value import CqlEmpty  # pyright: ignore[reportMissingModuleSource]
+
+import pytest
+import uuid
+import ipaddress
+import datetime
+import math
+from decimal import Decimal
+from datetime import time
+from dateutil.relativedelta import relativedelta
+import pytest_asyncio
+from scylla._rust.session import Session  # pyright: ignore[reportMissingModuleSource]
+from scylla._rust.session_builder import SessionBuilder  # pyright: ignore[reportMissingModuleSource]
+
+from scylla._rust.results import RowFactory, ColumnIterator  # pyright: ignore[reportMissingModuleSource]
+
+
+async def set_up() -> Session:
+    builder = SessionBuilder(["127.0.0.2"], 9042)
+    session = await builder.connect()
+
+    # 2. Create keyspace & table
+    await session.execute("""
+            CREATE KEYSPACE IF NOT EXISTS testks
+            WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1};
+        """)
+
+    await session.execute("USE testks")
+
+    return session
+
+
+@pytest_asyncio.fixture(scope="module")
+async def session():
+    session = await set_up()
+    yield session
+    await session.execute("DROP KEYSPACE testks")
+
+
+TableFactory = Callable[[str, str], Awaitable[str]]
+
+
+@pytest_asyncio.fixture
+async def table_factory(session: Session) -> AsyncGenerator[TableFactory, None]:
+    created_tables: list[str] = []
+
+    async def create_table(schema: str, name: str) -> str:
+        await session.execute(f"CREATE TABLE IF NOT EXISTS {name} ({schema});")
+        created_tables.append(name)
+        return name
+
+    yield create_table
+
+    for table in created_tables:
+        await session.execute(f"DROP TABLE IF EXISTS {table};")
+
+
+async def insert_and_fetch_single_row(
+    session: Session,
+    table_factory: TableFactory,
+    schema: str,
+    table_name: str,
+    row_id: int,
+    value_sql: str,
+) -> Dict[str, Any]:
+    table = await table_factory(schema, table_name)
+
+    await session.execute(f"INSERT INTO {table} (id, value) VALUES ({row_id}, {value_sql});")
+
+    result = await session.execute(f"SELECT * FROM {table} WHERE id = {row_id}")
+    return next(result.iter_rows())
+
+
+# Verifies that iter_rows() returns an iterator yielding row dictionaries
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_rows_result_is_iterator_and_dicts(session: Session, table_factory: TableFactory):
+    table = await table_factory(
+        "id int PRIMARY KEY, x int",
+        "rows_result_basic_table",
+    )
+
+    await session.execute(f"INSERT INTO {table} (id, x) VALUES (1, 10);")
+    await session.execute(f"INSERT INTO {table} (id, x) VALUES (2, 20);")
+
+    result = await session.execute(f"SELECT * FROM {table}")
+    rows = result.iter_rows()
+
+    # rows_result should be an iterator
+    assert iter(rows) is rows
+
+    # iterating should yield dicts
+    first = next(rows)
+    assert isinstance(first, dict)
+
+    assert "id" in first
+    assert "x" in first
+
+
+# Verifies correct deserialization of CQL text values, including Unicode
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+@pytest.mark.parametrize(
+    "row_id,value",
+    [(1, "hello"), (2, "Zażółć gęślą jaźń 🚀"), (3, "")],
+)
+async def test_text_deserialization(
+    session: Session,
+    table_factory: TableFactory,
+    row_id: int,
+    value: str,
+):
+    row = await insert_and_fetch_single_row(
+        session=session,
+        table_factory=table_factory,
+        schema="id int PRIMARY KEY, value text",
+        table_name="text_table",
+        row_id=row_id,
+        value_sql=f"'{value}'",
+    )
+
+    assert isinstance(row["value"], str)
+    assert row["value"] == value
+
+
+# Verifies correct deserialization of CQL float values
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+@pytest.mark.parametrize(
+    "row_id,value_sql,expected",
+    [
+        (1, "1.25", 1.25),
+        (2, "-42.5", -42.5),
+        (3, "3.234", 3.234),
+        (4, "Infinity", math.inf),
+        (5, "-Infinity", -math.inf),
+        (6, "NaN", math.nan),
+    ],
+)
+async def test_float_deserialization(
+    session: Session,
+    table_factory: TableFactory,
+    row_id: int,
+    value_sql: str,
+    expected: float,
+):
+    def to_float32(val: float) -> float:
+        import struct
+
+        # Pack as 32-bit float (f), then unpack back to Python float
+        return struct.unpack("f", struct.pack("f", val))[0]
+
+    row = await insert_and_fetch_single_row(
+        session=session,
+        table_factory=table_factory,
+        schema="id int PRIMARY KEY, value float",
+        table_name="float_table",
+        row_id=row_id,
+        value_sql=value_sql,
+    )
+
+    assert isinstance(row["value"], float)
+    if math.isnan(expected):
+        assert math.isnan(row["value"])
+    else:
+        assert row["value"] == to_float32(expected)
+
+
+# Verifies correct deserialization of CQL double values
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+@pytest.mark.parametrize(
+    "row_id,value",
+    [
+        (1, 3.141592653589793),
+        (2, -0.00000012345),
+    ],
+)
+async def test_double_deserialization(
+    session: Session,
+    table_factory: TableFactory,
+    row_id: int,
+    value: float,
+):
+    row = await insert_and_fetch_single_row(
+        session=session,
+        table_factory=table_factory,
+        schema="id int PRIMARY KEY, value double",
+        table_name="double_table",
+        row_id=row_id,
+        value_sql=str(value),
+    )
+
+    assert isinstance(row["value"], float)
+    assert row["value"] == value
+
+
+# Verifies correct deserialization of CQL list<int> into Python lists
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+@pytest.mark.parametrize(
+    "row_id,value",
+    [(1, [1, 2, 3]), (2, [35, 53, 15, 12]), (3, [])],
+)
+async def test_list_deserialization(
+    session: Session,
+    table_factory: TableFactory,
+    row_id: int,
+    value: list[int],
+):
+    row = await insert_and_fetch_single_row(
+        session=session,
+        table_factory=table_factory,
+        schema="id int PRIMARY KEY, value list<int>",
+        table_name="list_table",
+        row_id=row_id,
+        value_sql=str(value),
+    )
+
+    assert isinstance(row["value"], list)
+    assert row["value"] == value
+
+
+# Verifies correct deserialization of nested CQL list<frozen<list<int>>
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+@pytest.mark.parametrize(
+    "row_id,value",
+    [(1, [[1, 2, 3], [4, 5], [3, 2, 1]]), (2, [[10], [20, 30, 40], [3, 2, 1], [3, 2, 1]]), (3, [[]])],  # pyright: ignore
+)
+async def test_nested_list_deserialization(
+    session: Session,
+    table_factory: TableFactory,
+    row_id: int,
+    value: List[List[int]],
+):
+    row = await insert_and_fetch_single_row(
+        session=session,
+        table_factory=table_factory,
+        schema="id int PRIMARY KEY, value list<frozen<list<int>>>",
+        table_name="nested_list_table",
+        row_id=row_id,
+        value_sql=str(value),
+    )
+
+    nested = row["value"]
+
+    assert isinstance(nested, list)
+    for inner in nested:  # pyright: ignore[reportUnknownVariableType]
+        assert isinstance(inner, list)
+
+    assert nested == value
+
+
+# Verifies correct deserialization of CQL UDT into Python dictionaries
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+@pytest.mark.parametrize(
+    "row_id,value_sql,expected",
+    [
+        (
+            1,
+            "{ street: 'Main St', number: 10 }",
+            {"street": "Main St", "number": 10},
+        ),
+        (
+            2,
+            "{ street: 'Oak Ave', number: 42 }",
+            {"street": "Oak Ave", "number": 42},
+        ),
+    ],
+)
+async def test_udt_deserialization(
+    session: Session,
+    table_factory: TableFactory,
+    row_id: int,
+    value_sql: str,
+    expected: Dict[str, int],
+):
+    # Create UDT (safe to re-run)
+    await session.execute(
+        """
+        CREATE TYPE IF NOT EXISTS testks.address (
+            street text,
+            number int
+        )
+        """
+    )
+
+    row = await insert_and_fetch_single_row(
+        session=session,
+        table_factory=table_factory,
+        schema="id int PRIMARY KEY, value address",
+        table_name="udt_table",
+        row_id=row_id,
+        value_sql=value_sql,
+    )
+
+    assert isinstance(row["value"], dict)
+    assert row["value"] == expected
+
+
+# Verifies correct deserialization of list<frozen<UDT>> across multiple rows
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_list_udt_deserialization(session: Session, table_factory: TableFactory):
+    await session.execute("""
+        CREATE TYPE IF NOT EXISTS testks.address (
+            street text,
+            number int
+        );
+    """)
+
+    table = await table_factory("id int, name text, addrs list<frozen<address>>, PRIMARY KEY (id, name)", "nested_udt")
+
+    # 3. Insert multiple rows with list<udt>
+    rows_to_insert = 9  # 10–15 as requested
+    for i in range(rows_to_insert):
+        await session.execute(f"""
+            INSERT INTO {table} (id, name, addrs)
+            VALUES (
+                0,
+                'User{i}',
+                [
+                    {{ street: 'A-Street-{i}', number: {i * 10} }},
+                    {{ street: 'B-Street-{i}', number: {i * 10 + 1} }}
+                ]
+            );
+        """)
+
+    # 4. Query + deserialize
+    result = await session.execute(f"SELECT * FROM {table} WHERE id = 0 ORDER BY name ASC")
+    rows = result.iter_rows()
+    row_list = list(rows)
+
+    # 5. Assertions — verify all rows returned + structure is correct
+    assert len(row_list) == rows_to_insert
+
+    for i, row in enumerate(row_list):
+        assert isinstance(row["addrs"], list)
+        assert row["name"] == f"User{i}"
+        assert row["addrs"][0]["street"] == f"A-Street-{i}"
+        assert row["addrs"][0]["number"] == i * 10
+        assert row["addrs"][1]["street"] == f"B-Street-{i}"
+        assert row["addrs"][1]["number"] == i * 10 + 1
+
+
+# Verifies that a custom RowFactory can transform rows into user-defined Python objects
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_custom_row_factory_transforms_rows(session: Session, table_factory: TableFactory):
+    # Define a simple Python class for output rows
+    class UserRow:
+        def __init__(self, id: int, name: str, scores: list[int]):
+            self.id = id
+            self.name = name
+            self.scores = scores
+
+    # A valid RowFactory implementation
+    class UserFactory(RowFactory):
+        cls = UserRow
+
+        def build(self, column_iterator: ColumnIterator) -> Any:
+            id = 0
+            name = "Some"
+            scores = []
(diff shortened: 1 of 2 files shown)


src/value.rs
```diff
@@ -0,0 +1,170 @@
...
+// The trait does not provide a `Python` argument, meaning that Python objects which
+// would have to be constructed inside `deserialize()` or deeper in recursion
+// would require acquiring the GIL separately for every column/element/UDT/List
+// during nested decoding.
+//
+// Using our own `PyDeserializeValue` trait keeps all Python conversions inside
+// a single `Python::with_gil(...)` boundary, allowing complex values (lists,
+// sets, UDTs, nested collections) to be deserialized without repeated GIL
+// acquisition. This avoids potential slowdown of acquiring GIL multiples times.
+//
```
Lorak-mmk: Note: We wouldn't neet to acquire and release GIL for each value.
We would need to acquire it once, and then in each value check that it is already acquired.

https://pyo3.rs/v0.27.1/performance.html#access-to-bound-implies-access-to-python-token

So the reasoning is fine, but the comment is a bit imprecise.

wprzytula:
@patrycja-ziemkiewicz Please rebase on newest main, then reask for review.

src/value.rs
```diff
@@ -0,0 +1,181 @@
...
+    ) -> Result<PyDeserializedValue<'py>, DeserializationError> {
+        let elem_typ = match typ {
+            ColumnType::Collection {
+                frozen: false,
+                typ: CollectionType::List(elem_typ),
+            }
+            | ColumnType::Collection {
+                frozen: false,
+                typ: CollectionType::Set(elem_typ),
+            } => elem_typ,
```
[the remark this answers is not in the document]
wprzytula:
```suggestion
        let elem_typ = match typ {
            ColumnType::Collection {
                frozen: _,
                typ: CollectionType::List(elem_typ),
            }
            | ColumnType::Collection {
                frozen: _,
                typ: CollectionType::Set(elem_typ),
            } => elem_typ,
```

src/value.rs
```diff
@@ -0,0 +1,181 @@
...
+
+impl<'frame, 'metadata, 'py, T> PyDeserializeValue<'frame, 'metadata, 'py> for Vec<T>
+where
+    T: PyDeserializeValue<'frame, 'metadata, 'py>,
+{
+    fn deserialize_py(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+        py: Python<'py>,
+    ) -> Result<PyDeserializedValue<'py>, DeserializationError> {
```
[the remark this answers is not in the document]
wprzytula: So this `Vec` serves just for static typing purposes, before we type-erase anyway into `PyAny`. Correct?

src/value.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+impl<'py> IntoPyObject<'py> for PyDeserializedValue<'py> {
+    type Target = PyAny;
+    type Output = Bound<'py, PyAny>;
+    type Error = Infallible;
+    fn into_pyobject(self, _: Python<'py>) -> Result<Self::Output, Self::Error> {
+        Ok(self.value)
+    }
+}
```
wprzytula: :pick: missing empty line between `impl` blocks.

src/value.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn deserialize_py(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+        py: Python<'py>,
+    ) -> Result<Self, DeserializationError> {
+        let cql = deser_cql_py_value(py, typ, v)?;
+
+        Ok(PyDeserializedValue { value: cql })
+    }
+}
```
wprzytula: :bulb: A-ha! So this is the main point of interest to us. All implementation on collections that I saw above are just to be used in `deser_cql_py_value`. Neat and reasonable :+1:

src/value.rs
```diff
@@ -0,0 +1,181 @@
...
+                Ok(PyInt::new(py, v).into_any())
+            }
+            NativeType::Text => {
+                let v = <&str as DeserializeValue<'frame, 'metadata>>::deserialize(typ, val)?;
+                Ok(PyString::new(py, v).into_any())
+            }
+            NativeType::Boolean => {
+                let v = bool::deserialize(typ, val)?;
+                Ok(PyBool::new(py, v).to_owned().into_any())
+            }
```
wprzytula: :question: What does this `to_owned()` do here? Why is this needed specifically for boolean, while not needed for other primitive types?

src/value.rs
```diff
@@ -0,0 +1,181 @@
...
+            _ => unimplemented!(),
+        },
+        ColumnType::Collection {
+            frozen: _frozen,
+            typ: col_typ,
+        } => match col_typ {
+            CollectionType::List(_type_name) => {
+                let list = Vec::<PyDeserializedValue>::deserialize_py(typ, val, py)?;
+                list.into_pyobject(py).map_err(DeserializationError::new)
+            }
```
wprzytula: :thought_balloon: While this is still quite strange for me to see `Vec::deserialize_py` not deserialize to `Vec`, I really appreciate the tidyness of this solution. :100:

src/results.rs
```diff
@@ -0,0 +1,278 @@
...
+use stable_deref_trait::StableDeref;
+use std::fmt::Write;
+use std::ops::Deref;
+use yoke::{Yoke, Yokeable};
+
+// Originally written by @Lorak-mmk, moved by @patrycja-ziemkiewicz
+#[pyclass]
+pub(crate) struct RequestResult {
+    pub(crate) inner: scylla::response::query_result::QueryResult,
+}
```
wprzytula: :wrench: Let's make sure this commit has `Co-authored-by` mention.

src/results.rs
```diff
@@ -0,0 +1,278 @@
...
+                match col {
+                    Some(c) => write!(result, "{}", c).unwrap(),
+                    None => write!(result, "null").unwrap(),
+                };
+                write!(result, "|").unwrap();
+            }
+            writeln!(result).unwrap();
+        }
+        Ok(PyString::new(py, &result))
+    }
```
wprzytula: :open_mouth:  Wow, is this essentially a query result displayer? Do we want such thing? @Lorak-mmk

src/results.rs
```diff
@@ -0,0 +1,278 @@
...
+            row_col_iterator,
+            factory: f,
+        })
+    }
+}
+#[pyclass]
+pub struct RowsResult {
+    row_col_iterator: Py<RowColumnIterator>,
+    factory: Py<RowFactory>,
+}
```
wprzytula: :pick: Missing empty line.

src/results.rs
```diff
@@ -0,0 +1,278 @@
...
+        slf
+    }
+}
+
+#[derive(Yokeable)]
+struct IteratorWrapper<'a> {
+    row_iterator: RawRowIterator<'a, 'a>,
+    column_iterator: ColumnIterator<'a, 'a>,
+}
+impl<'a> IteratorWrapper<'a> {
```
wprzytula: :pick: Missing empty line.

src/results.rs
```diff
@@ -0,0 +1,278 @@
...
+
+// NOTE: In method attach_to_cart we cannot move out cart so We cannot use into_inner inside of it.
+// Fields in DeserializedMetadataAndRawRows are private We have to think
+// If it would be better to use method into_inner and construct new RawMetadata struct
+// or to implement some public method in rust driver that would give access to raw_rows.
+struct RawMetadata {
+    metadata: ResultMetadataHolder,
+    rows_count: usize,
+    raw_rows: Bytes,
+}
```
wprzytula: :+1: `raw_rows` getter has been recently exposed for Node.js-RS Driver. I'm sure we can expose it for Python-RS as well.

src/results.rs
```diff
@@ -0,0 +1,278 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+unsafe impl StableDeref for RawMetadataCart {}
+
+#[pyclass(name = "ColumnIterator")]
+pub struct RowColumnIterator {
+    yoked: Yoke<IteratorWrapper<'static>, RawMetadataCart>,
+}
```
wprzytula: :question: Why is this called "RowColumnIterator"? What's a "RowColumn"? How do I parse this?

src/results.rs
```diff
@@ -0,0 +1,278 @@
...
+        slf
+    }
+}
+
+#[pyclass]
+pub struct Column {
+    #[pyo3(get)]
+    column_name: Py<PyString>,
+    #[pyo3(get)]
+    value: Py<PyAny>,
```
wprzytula: :question: Can we make this typed? Don't we have a union type for all possible Python values deserializable from CQL?

python/tests/test_deserialization.py
```diff
@@ -0,0 +1,279 @@
...
+async def test_simple_deserialization():
+    # 1. Connect
+    builder = SessionBuilder(["127.0.0.2"], 9042)
+    session = await builder.connect()
+
+    # 2. Create keyspace & table
+    await session.execute("""
+        CREATE KEYSPACE IF NOT EXISTS testks
+        WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};
+    """)
```
wprzytula: :wrench: Please never use `SimpleStrategy` - just replace it with `NetworkTopologyStrategy`.

python/tests/test_deserialization.py
```diff
@@ -0,0 +1,279 @@
...
+    session = await builder.connect()
+
+    # 3. Create a table with complex CQL types
+    await session.execute("""
+                          CREATE TABLE IF NOT EXISTS testks.complex_table2 (
+                                                                              id int PRIMARY KEY,
+                                                                              name text,
+                                                                              scores list<int>,
+                          );
+                          """)
```
wprzytula: :wrench: Can we make this test use nested lists? Just to have this additional case tested.

wprzytula (commented):
:100: Looks very promising - great work! :rocket:

src/value.rs
```diff
@@ -0,0 +1,170 @@
...
+
+            let dict = PyDict::new(py);
+
+            for ((col_name, col_type), res) in iter {
+                let val = res.and_then(|v| {
+                    Option::<PyCqlValue>::deserialize_py(col_type, v.flatten(), py)
+                })?;
+
+                dict.set_item(col_name.to_string(), val).map_err(DeserializationError::new)?;
+            }
```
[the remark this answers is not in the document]
patrycja-ziemkiewicz: Even though key/value conversion is infallible here, PyDict_SetItem can still fail at the Python runtime level.

src/value.rs
```diff
@@ -0,0 +1,763 @@
...
+    fn deserialize_py(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+        py: Python<'py>,
+    ) -> Result<PyDeserializedValue, DeserializationError>;
+}
+
+pub struct List<T> {
+    phantom_data: std::marker::PhantomData<T>,
+}
```
Lorak-mmk: Let's keep `pub` used only for items that are exposed to Python, and `pub(crate)` for items needed outside the module defining them. In other words, privacy of items should be as narrow as possible.
This is a general note, not only for those 2 items - I see this issue in other places, so please review your code and adjust.

src/value.rs
Lorak-mmk: Commit: "Implement value deserialization logic"

In this commit `value.rs` is not included in the module tree (it is not defined in `lib.rs`), so its effectively a dead code. You could just fix it by including it here, but even then it will be effectively dead code. It seems to me that a better way to split this PR into commits would be the following:
- First perform the changes to `execute`, `RequestResult` etc. In other words, introduce the deserialization framework. This commit would include the traits, and other necessary things, but it would not include implementations of the trait, apart from let's say implementation for an int.
- Second commit adds all the other implementations, and expands type hints to include them.
- Third commit adds tests.

This would get rid of any dead code, and would make it easier to review the code: reader could immediately see how the traits you introduce are going to be used.

src/value.rs
```diff
@@ -0,0 +1,763 @@
...
+                let v = Counter::deserialize(typ, val)?;
+                Ok(PyInt::new(py, v.0).into_any())
+            }
+            NativeType::Decimal => {
+                let dec = CqlDecimal::deserialize(typ, val)?;
+                let big: bigdecimal_04::BigDecimal = dec.into();
+                call_python_constructor(py, "decimal", "Decimal", (big.to_string(),), None)
+                    .map_err(DeserializationError::new)
+            }
+            NativeType::TinyInt => {
```
Lorak-mmk: Could you check what python-driver and scyllapy deserialize cql Decimal to? Do they also use `decimal.Decimal`?

src/value.rs
```diff
@@ -0,0 +1,763 @@
...
+                py_call_method(
+                    py,
+                    "datetime",
+                    "datetime",
+                    "fromtimestamp",
+                    (seconds,),
+                    Some(&kwargs),
+                )
+                .map_err(DeserializationError::new)
+            }
```
Lorak-mmk: This also look unnecessarily complicated. PyO3 has support for chrono and time - can't we use one of those here?

src/value.rs
```diff
@@ -0,0 +1,763 @@
...
+                let time: time_03::Time = t.try_into().map_err(DeserializationError::new)?;
+
+                let hour = time.hour();
+                let minute = time.minute();
+                let second = time.second();
+                let micro = time.nanosecond() / 1000;
+
+                call_python_constructor(py, "datetime", "time", (hour, minute, second, micro), None)
+                    .map_err(DeserializationError::new)
+            }
```
Lorak-mmk: ditto

src/value.rs
```diff
@@ -0,0 +1,763 @@
...
+            NativeType::Uuid => {
+                let v = uuid::Uuid::deserialize(typ, val)?;
+                call_python_constructor(py, "uuid", "UUID", (v.to_string(),), None)
+                    .map_err(DeserializationError::new)
+            }
+            NativeType::Timeuuid => {
+                let v = CqlTimeuuid::deserialize(typ, val)?;
+                call_python_constructor(py, "uuid", "UUID", (v.to_string(),), None)
+                    .map_err(DeserializationError::new)
+            }
```
Lorak-mmk: Also please verify that python-driver and scyllapy use uuid for timeuuid.

src/value.rs
Lorak-mmk: What's the purpose of this type? Can't you implement your trait on PyList?

src/value.rs
wprzytula: This is OK IMO then.

src/value.rs
```diff
@@ -0,0 +1,763 @@
...
+
+                let time: time_03::Time = t.try_into().map_err(DeserializationError::new)?;
+
+                let hour = time.hour();
+                let minute = time.minute();
+                let second = time.second();
+                let micro = time.nanosecond() / 1000;
+
+                call_python_constructor(py, "datetime", "time", (hour, minute, second, micro), None)
+                    .map_err(DeserializationError::new)
```
wprzytula: ♻️ Could we maybe extract branches of this huge match to distinct functions? So that everything would get more modular.

wprzytula: But it does make sense IMO to keep dispatcher (this function with the large `match`) separate from actual implementation. Jumpling between functions is only needed if you need to really read a particular implementation; what usually matters is ease of grasping the code top-down, which is easier if your functions are shorter and delegate things to intuitively extracted and named subroutines.

This is just my opinion, but I feel confident about it.

src/value.rs
```diff
@@ -0,0 +1,763 @@
...
+                    .map_err(DeserializationError::new)?;
+
+                call_python_constructor(
+                    py,
+                    "dateutil.relativedelta",
+                    "relativedelta",
+                    (),
+                    Some(&kwargs),
+                )
+                .map_err(DeserializationError::new)
```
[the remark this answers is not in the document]
wprzytula: Perhaps it is. Here we handle errors differently (by returning custom DeserializationError) than we would in IntoPyObject impl AFAIU.

src/value.rs
patrycja-ziemkiewicz: They both use uuid.UUID.

python/scylla/_rust/results.pyi
```diff
@@ -0,0 +1,118 @@
...
+    # - Vector
+    List["CqlValue"],
+
+    # CQL:
+    # - Set
+    Set["CqlValue"],
+
+    # CQL:
+    # - Tuple
+    Tuple["CqlValue", ...],
```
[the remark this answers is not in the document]
patrycja-ziemkiewicz: In legacy python driver it is deserialized to list.

src/deserialize/value.rs
```diff
@@ -0,0 +1,721 @@
...
+                let Ok(obj) = map.into_pyobject(py);
+                Ok(obj)
+            }
+            // CQL Set → Python set
+            CollectionType::Set(_type_name) => {
+                let set = Set::<PyDeserializedValue>::deserialize_py(typ, Some(val), py)?;
+                let Ok(obj) = set.into_pyobject(py);
+                Ok(obj)
+            }
+            _ => unimplemented!(),
```
wprzytula: :wrench: Let's provide more context to this `unimplemented()` message.

src/deserialize/value.rs
```diff
@@ -0,0 +1,721 @@
...
+                        Some(v) => PyDeserializedValue::deserialize_py(typ, Some(v), py),
+                        None => Ok(PyDeserializedValue::new(py_none(py))),
+                    });
+                PyValueOrError::new(result)
+            });
+
+            let tuple = PyTuple::new(py, t)?;
+            Ok(tuple.into_any())
+        }
+        _ => unimplemented!(),
```
wprzytula: Ditto: more context needed in `unimplemented!()`.

src/deserialize/conversion.rs
```diff
@@ -0,0 +1,216 @@
...
+            add_base10(&mut output, &temp);
+        }
+
+        // base *= 256 for the next byte
+        mul_scalar_base10(&mut base, 256);
+    }
+
+    output.reverse();
+    output
+}
```
wprzytula: :warning: This allocates one `Vec` per each non-zero digit... This is horrible!

wprzytula (changes requested):

src/deserialize/conversion.rs
patrycja-ziemkiewicz: I removed the per-digit Vec allocation, but as mentioned earlier this implementation is not optimal and can be further improved.

python/scylla/_rust/results.pyi
```diff
@@ -0,0 +1,103 @@
...
+    # CQL:
+    # - Time
+    time,
+    # CQL:
+    # - Duration
+    relativedelta,
+    # CQL:
+    # - Empty
+    # - null
+    None,
```
Lorak-mmk: If this is only used for deserialization, then Empty can be dropped from the comment I think.
If this is used also for serialization, then we need a separate type for Empty.

python/tests/test_deserialization.py
```diff
@@ -0,0 +1,1060 @@
...
+
+    for inner in nested1:  # pyright: ignore[reportUnknownVariableType]
+        assert isinstance(inner, list)
+
+    for inner in nested2:  # pyright: ignore[reportUnknownVariableType]
+        assert isinstance(inner, list)
+
+    # Value checks
+    assert nested1 == v1
+    assert nested2 == v2
```
Lorak-mmk: Please add a case with empty inner list.

python/tests/test_deserialization.py
```diff
@@ -0,0 +1,1060 @@
...
+        (5, Decimal("-7890.00123")),
+        (6, Decimal("123456789012345678901234567890.123456789")),
+        (7, Decimal("-999999999999999999999999999999999999.9999")),
+        (8, Decimal("0.00000000000000000001")),
+        (9, Decimal("42.000000")),
+        (10, Decimal("617283694")),
+        (11, Decimal("617283694.4324")),
+    ],
+)
+async def test_decimal_deserialization(
```
Lorak-mmk: You wrote this test using parametrize, but in all other tests you duplicated all test lines for all cases.

Please remove the duplication in other tests. We don't want to have to duplicate whole test of e.g. int just to add a test case.
Wheter you use `parametrize` or just an array defined inside test is up to you, I don't have enough knowledge about Python testing conventions to decide that.

python/tests/test_deserialization.py
```diff
@@ -0,0 +1,1060 @@
...
+    await session.execute(f"INSERT INTO {table} (id, ts) VALUES (2, {ts2});")
+
+    rows = list((await session.execute(f"SELECT * FROM {table}")).iter_rows())
+
+    r1 = next(r for r in rows if r["id"] == 1)
+    r2 = next(r for r in rows if r["id"] == 2)
+
+    assert r1["ts"] == dt1
+    assert r2["ts"] == dt2
+
```
Lorak-mmk: Again, please introduce regression test for edge cases. This comment applies also to other date/time related types where you used fallible conversion in deserialization.

src/deserialize/results.rs
```diff
@@ -159,29 +151,31 @@ unsafe impl StableDeref for DeserializedMetadataAndRawRowsCart {}
...
-        let cart = DeserializedMetadataAndRawRowsCart(Box::new(data));
-
-        let yoked = Yoke::attach_to_cart(cart, |cart| {
-            let frame_slice = FrameSlice::new(cart.raw_rows());
+    fn new(query_result: Arc<QueryResult>) -> PyResult<Self> {
+        let cart = QueryResultCart(query_result);
+
+        let yoked = Yoke::try_attach_to_cart(cart, |cart| -> PyResult<_> {
+            let Some(raw_rows_with_metadata) = cart.deserialized_metadata_and_rows() else {
+                return Err(PyRuntimeError::new_err("Result does not have rows"));
```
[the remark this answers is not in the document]
wprzytula: IMO iterating over non-rows result should be possible - such iteration should return None immediately. This is because we're in Python, where (I believe) we care mostly about ease of use, less about rigidity.

python/scylla/_rust/results.pyi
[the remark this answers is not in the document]
wprzytula: But in Python there's no distinction tuple vs list in the aspect of heterogeneity. The main (maybe the only one) difference is in mutability, and I see Vector as a fixed length immutable type, which matches Python tuple.

src/deserialize/value.rs
```diff
@@ -91,13 +516,190 @@ fn deser_cql_py_value<'py, 'metadata, 'frame>(
...
+                Ok(PyBool::new(py, v).to_owned().into_any())
+            }
+            // CQL Date → Python datetime.date
+            NativeType::Date => {
+                let date: chrono_04::NaiveDate = CqlDate::deserialize(typ, Some(val))?
+                    .try_into()
+                    .map_err(DeserializationError::new)?;
+
+                Ok(date.into_pyobject(py)?.into_any())
+            }
```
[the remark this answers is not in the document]
wprzytula: +1 for opening the issue.

src/deserialize/conversion.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    fn into_pyobject(self, py: Python<'py>) -> Result<Self::Output, Self::Error> {
+        let bytes = self.val.as_signed_bytes_be_slice();
+        unsafe {
+            let val = ffi::_PyLong_FromByteArray(bytes.as_ptr(), bytes.len(), 0, 1);
+
+            Ok(Bound::from_owned_ptr(py, val).downcast_into()?)
+        }
+    }
+}
```
wprzytula: :question: Where and how did you find this `ffi::_PyLong_FromByteArray`? I had trouble when perusing PyO3 docs.

src/value.rs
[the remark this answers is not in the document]
wprzytula: It looks very good now.

src/value.rs
wprzytula: Now the branches are quite concise, so it's no longer relevant.

src/deserialize/value.rs
```diff
@@ -0,0 +1,763 @@
...
+                kwargs
+                    .set_item("months", d.months)
+                    .map_err(DeserializationError::new)?;
+                kwargs
+                    .set_item("days", d.days)
+                    .map_err(DeserializationError::new)?;
+                kwargs
+                    .set_item("microseconds", d.nanoseconds / 1000)
+                    .map_err(DeserializationError::new)?;
+
```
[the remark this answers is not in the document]
wprzytula: @Lorak-mmk WDYT?

src/deserialize/results.rs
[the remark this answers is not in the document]
wprzytula: Please check what old Python driver does.

python/scylla/_rust/results.pyi
Lorak-mmk: The database rejects empty values send by the driver?

wprzytula: OK, this is indeed supported for `CqlValue::Empty`. So `CqlValue` is the only way to insert empty CQL values.

wprzytula: We should probably implement `SerializeValue` for `MaybeEmptiable` then.

Issue #11, opened 2025-09-15
Support prepared statements


Lorak-mmk:
What we need:
- async `prepare` method on `Session`, which returns `PreparedStatement` object.
- `execute` method should be able to accept either string (as it does now) or such `PreparedStatement`.

To write a type hint for `execute` method after this change we can use either `Union` (because for now only first argument differs), or overload ( https://typing.python.org/en/latest/spec/overload.html ).
I don't think we plan to have a difference in behavior between prepared and unrprepared statements, so `Union` is fine.

Alternative approaches to API:
- We could to what gocql does and auto-prepare statements, similar to what CachingSession does in Rust Driver. I don't think this is a good default. I'm not sure DML statements (CREATE KEYSPACE / TABLE) etc can even be prepared. Some statements should not be prepared (for example if they don't specify keyspace explicitly). 
- We could have different methods for executing prepared and unprepared statements, as we do in Rust Driver. Even in Rust we have some plans to unify the methods, and given Pythons dynamic nature, overloaded method seems like a good fit. It is also more similar to the old driver, which will be an advantage for migration.

PR #26, opened 2025-11-28
Prepared statements

jakub-graczyk:
Fixes: #11 
Added `PreparedStatement` and `Session.prepare()`.
Updated `execute` to use `PreparedStatement` or python str.

Added tests for `execute` with `PreparedStatement`.

### DIFF
--- a/python/tests/test_prepare.py
+++ b/python/tests/test_prepare.py
@@ -1,0 +1,35 @@
+import pytest
+from scylla.session_builder import SessionBuilder
+from scylla.statement import PreparedStatement
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_prepare_statement():
+    builder = SessionBuilder(["127.0.0.2"], 9042)
+    session = await builder.connect()
+    prepared = await session.prepare("SELECT * FROM system.local")
+    print(prepared)
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_prepare_and_execute():
+    builder = SessionBuilder(["127.0.0.2"], 9042)
+    session = await builder.connect()
+    prepared = await session.prepare("SELECT * FROM system.local")
+    assert isinstance(prepared, PreparedStatement)
+    result = await session.execute(prepared)
+    print(result)
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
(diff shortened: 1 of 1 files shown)


src/session.rs
```diff
@@ -2,11 +2,32 @@ use std::fmt::Write;
...
+                    .map_err(|e| PyRuntimeError::new_err(
+                        format!("Failed to deserialize metadata: {}", e)
+                    ))
+            })
+            .await
+            .expect("Driver should not panic")?;
+
+        return Ok(RequestResult { inner: result });
+    }};
+}
```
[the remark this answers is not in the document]
Lorak-mmk: It's fine to have a method on our session that hides the boilerplate of `Arc::clone` and `RUNTIME.spawn` by taking an async closure callback.

src/session.rs
```diff
@@ -15,22 +36,37 @@ pub(crate) struct Session {
...
+        let session_clone = std::sync::Arc::clone(&self._inner);
+        match session_clone.prepare(statement).await {
+            Ok(prepared) => Ok(PreparedStatement {
+                _inner: Arc::new(prepared),
+            }),
+            Err(e) => Err(PyErr::new::<PyRuntimeError, _>(format!(
+                "Failed to prepare statement: {}",
+                e
+            ))),
+        }
```
Lorak-mmk: Whoa, you are not using `RUNTIME` here. Is that code working? How is that possible?

Lorak-mmk (changes requested):

src/session.rs
jakub-graczyk: My intuition is: in Rust Driver when calling `session.query()` you eventually call `run_request` which uses `tokio::time::timeout()` and that's why we need to spawn it in runtime. I don't see tokio calls with `session.prepare()` so there is no need for runtime? Remember with benchmarks regular async functions could be called with with pyo3 w/o runtime.

Lorak-mmk: When you were making benchmarks, did you also benchmark this case (calling async Rust functions from Python, without using runtime)?
Maybe `prepare` does not use tokio directly, but it surely uses a channel (to send a request to connection). Maybe a more performant solution for us to perform request is to do the same: have a global task to which we will send requests (and this task will call execute etc), avoiding the need for spawn.

PR #27, opened 2025-11-30
Implemented deserialization CQL -> Rust -> Python + test

pasinskik:
Implemented CqlValue-based row deserialization in Rust and Python type mapping.

- Added RustCqlRow (DeserializeRow implementation returning CqlValue)
- Added cql_value_to_py mapper (CqlValue -> Python objects)
- Added RequestResult.rows_as_dicts()
- Ensured support of primitive types, lists, maps, UDTs and inet
- Provided minimal integration test

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -75,22 +75,22 @@ impl RequestResult {
         Ok(PyString::new(py, &result))
     }
 
-    // Convert all rows to a list of dictionaries
-    pub fn rows_as_dicts(&self, py: Python<'_>) -> PyResult<Vec<PyObject>> {
-        // Convert the QueryResult into RowsResult
+    // Convert all rows to a Python list of dictionaries
+    pub fn rows_as_dicts(&self, py: Python<'_>) -> PyResult<PyObject> {
         let rows_result = self
             .inner
             .clone()
             .into_rows_result()
             .map_err(|e| PyRuntimeError::new_err(format!("non-rows result: {e}")))?;
 
-        // Iterate over the rows and convert each to a dictionary
+        // Iterate over the rows and onvert each to RustCqlRow
         let rows_iter = rows_result
             .rows::<RustCqlRow>()
             .map_err(|e| PyRuntimeError::new_err(format!("Failed to deserialize rows: {e}")))?;
 
-        let mut out: Vec<PyObject> = Vec::new();
+        let py_list = pyo3::types::PyList::empty(py);
 
+        // For each row, convert to a Python dict and append to the list
         for row_res in rows_iter {
             let row = row_res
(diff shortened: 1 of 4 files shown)


src/session.rs
```diff
@@ -71,6 +74,45 @@ impl RequestResult {
...
+            for (name, opt_val) in row.columns {
+                let py_val = match opt_val {
+                    Some(ref cql) => cql_value_to_py(py, cql)?,
+                    None => py.None(),
+                };
+
+                dict.set_item(name, py_val).map_err(|e| {
+                    PyRuntimeError::new_err(format!("Failed to set dict item: {e}"))
+                })?;
+            }
```
Lorak-mmk: Only now you are converting to dict.

src/cqlvalue_to_py.rs
```diff
@@ -0,0 +1,79 @@
...
+        Float(f) => PyFloat::new(py, *f as f64).into_any(),
+
+        // Blob -> bytes
+        Blob(bytes) => PyBytes::new(py, bytes).into_any(),
+
+        // UUID
+        Uuid(u) => {
+            let uuid_mod = py.import("uuid")?;
+            let uuid_cls = uuid_mod.getattr("UUID")?;
+            uuid_cls.call1((u.to_string(),))?.into_any()
```
Lorak-mmk: How does Python imports work? Specifically, what is the performance of re-importing modules?
You will call this import and getattr for each deserialize, so if it is not optmized well in Python / pyo3, we may need to cache `uuid_mod` / `uuid_cls`, for example using https://doc.rust-lang.org/std/sync/struct.OnceLock.html

src/session.rs
```diff
@@ -71,6 +74,45 @@ impl RequestResult {
         }
         Ok(PyString::new(py, &result))
     }
+
+    // Convert all rows to a list of dictionaries
+    pub fn rows_as_dicts(&self, py: Python<'_>) -> PyResult<Vec<PyObject>> {
+        // Convert the QueryResult into RowsResult
+        let rows_result = self
```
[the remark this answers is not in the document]
Lorak-mmk: Commit: "Implemented deserialization CQL -> Rust -> Python + test "

I see this is not solved despite the comment being marked is solved.

Lorak-mmk (changes requested):
Ah, I see what happened. You added a second commit that fixes issues that I pointed out before. Please modify the original commit instead, otherwise you force the reviewer to read a code that is wrong, and not present in the final version (because it is fixed by subsequent commits).

wprzytula:
As discussed during the meeting, this is a duplicate of #25. I suggest closing this. @Lorak-mmk @pasinskik WDYT?

Issue #10, opened 2025-09-15
Implement execution profiles


Lorak-mmk:
Execution Profiles are bundles of configuration options influencing request execution. Such options include client-side timeout, consistency, load balancing policy, retry policy, and others.
Typically, there is a hierarchy of configuration options that the driver needs to resolve. For example, there can be a default execution profile set on a session, a statement can have its own execution profile set, and some options may be possible to overwrite without using execution statement (either on statement level, or single request level).

The task here has 3 main steps:
- Research how execution profiles and other request configuration is done in other drivers. We definitely need to know about Rust Driver, but it is also worth researching Java Driver (both 3.x and 4.x) because it is a very popular one ( https://github.com/scylladb/java-driver/ ), and Python Driver ( https://github.com/scylladb/python-driver ) because that is what we are replacing. It would be good to write it down here for others to learn, and facilitate discussion.
- With the knowledge above, propose an API that we should implement here. Describe it here, including the hierarchy and all the options we ultimately want to support. Key points:
   - It should be ergonomic, intuitive and idiomatic for Python programmers
   - It needs to be possible to implement using Rust Driver API. Preferably without rewriting too much stuff (like resolving final options from hierarchy).
   - It should encourage writing correct code and not have dangerous footguns.
- Implement it. For now we'll skip things like policies because they are not implemented yet.

jakub-graczyk:
# Java Driver 4.x

## Overview

The Java Driver 4.x uses `reference.conf` configuration file to control its behavior.

You can create your own configuration files to customize the driver's behavior.

## Default Configuration

The driver has a `reference.conf` file containing all default settings.

## Creating Execution Profile with config

You can override defaults by creating an `application.conf` file in your application's resources folder:

```/docs/source/manual/core/configuration/README.md
datastax-java-driver {
  profiles {
    oltp {
      basic.request.timeout = 100 milliseconds
      basic.request.consistency = ONE
    }
    olap {
      basic.request.timeout = 5 seconds
      basic.request.consistency = QUORUM
    }
}
```

## Creating Execution Profile programatically

```java
DriverExecutionProfile defaultProfile = session.getContext().getConfig().getDefaultProfile();
DriverExecutionProfile dynamicProfile =
  defaultProfile.withString(
      DefaultDriverOption.REQUEST_CONSISTENCY, DefaultConsistencyLevel.EACH_QUORUM.name());
SimpleStatement s =
[...]

PR #30, opened 2025-12-09
Add execution profiles

jakub-graczyk:
This PR aims to implement design discussed in #10.

Added `ExecutionProfile` (not including policies)

When using `SessionBuilder` additional argument `execution_profile=` can be specified.

Added methods on `PreparedStatement` to return new `PreparedStatement` objects with specified options.

### DIFF
--- a/python/tests/test_execution_profile.py
+++ b/python/tests/test_execution_profile.py
@@ -2,6 +2,8 @@ import pytest
 from scylla.enums import Consistency, SerialConsistency
 from scylla.execution_profile import ExecutionProfile
 from scylla.session_builder import SessionBuilder
+from scylla.statement import PreparedStatement
+from scylla.types import Unset
 
 
 def test_execution_profile_builder():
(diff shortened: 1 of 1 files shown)


python/scylla/_rust/enums.pyi
```diff
@@ -0,0 +1,7 @@
+class Consistency:
+    def __init__(self, consistency: str) -> None: ...
+    def __str__(self) -> str: ...
+
+class SerialConsistency:
+    def __init__(self, serial_consistency: str) -> None: ...
+    def __str__(self) -> str: ...
```
Lorak-mmk: ditto

python/scylla/_rust/execution_profile.pyi
```diff
@@ -0,0 +1,12 @@
...
+class ExecutionProfile:
+    def __init__(
+        self,
+        timeout: float | None = None,
+        consistency: Consistency | None = None,
+        serial_consistency: SerialConsistency | None = None,
+    ) -> None: ...
+    def get_request_timeout(self) -> float | None: ...
+    def get_consistency(self) -> Consistency: ...
+    def get_serial_consistency(self) -> SerialConsistency | None: ...
```
Lorak-mmk: @wprzytula Please review this part carefully. I see that `get_consistency` return non-optional consistency, but `get_serial_consistency` returns optional serial consistency. I remember there were some reasons for similar weirdness in Rust Driver, but I don't remember enough to correctly asses this code.

src/execution_profile.rs
```diff
@@ -0,0 +1,74 @@
...
+        if let Some(timeout) = timeout {
+            let secs = Python::with_gil(|py| timeout.extract::<f64>(py))?;
+            if !secs.is_finite() || secs <= 0.0 {
+                return Err(pyo3::exceptions::PyValueError::new_err(
+                    "timeout must be a positive, finite number (in seconds)",
+                ));
+            }
+            let millis = (secs * 1000.0) as u64;
+            profile_builder = profile_builder.request_timeout(Some(Duration::from_millis(millis)));
+        }
```
Lorak-mmk: nit: There is `Duration::from_secs_f32`. Btw is that the idiomatic way to pass durations in Python? Is ther nothing like a duration type? It's not obvious to me as a user if the float represents seconds, or milliseconds.

python/scylla/_rust/statement.pyi
```diff
@@ -1,2 +1,16 @@
...
+    def get_execution_profile(self) -> ExecutionProfile | None: ...
+    def set_consistency(self, c: Consistency) -> PreparedStatement: ...
+    def unset_consistency(self) -> PreparedStatement: ...
+    def get_consistency(self) -> Consistency | None: ...
+    def set_serial_consistency(self, sc: SerialConsistency) -> PreparedStatement: ...
+    def unset_serial_consistency(self) -> PreparedStatement: ...
+    def get_serial_consistency(self) -> SerialConsistency | None: ...
+    def set_request_timeout(self, timeout: float) -> PreparedStatement: ...
+    def unset_request_timeout(self) -> PreparedStatement: ...
+    def get_request_timeout(self) -> float | None: ...
```
Lorak-mmk: It is surprising that those methods return new statement, instead of modifying it. The `set/unset` terminology doesn't fit that approach.
- Why did you chose that approach?
- If we go with that approach, dont you think that `with/without` is better terminology than `set/unset`?

src/execution_profile.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub(crate) fn new(
+        timeout: Option<PyObject>,
+        consistency: Option<PyObject>,
+        serial_consistency: Option<PyObject>,
+    ) -> PyResult<Self> {
+        let mut profile_builder = client::execution_profile::ExecutionProfile::builder();
+
+        if let Some(timeout) = timeout {
+            let secs = Python::with_gil(|py| timeout.extract::<f64>(py))?;
+            if !secs.is_finite() || secs <= 0.0 {
```
Lorak-mmk: Note: We really need a good solution for error handling. This `?` here is most likely not what we will ultimately want (but for this PR its ok). Intead we should give users an error that explains which parameter has wrong type.

python/tests/test_execution_profile.py
```diff
@@ -54,3 +55,15 @@ def test_execution_profile_timeout():
...
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_create_session_with_profile():
+    expected_timeout = 10.5
+    expected_consistency = Consistency.All
+    profile = ExecutionProfile(timeout=expected_timeout, consistency=expected_consistency)
+    builder = SessionBuilder(["127.0.0.2"], 9042, execution_profile=profile)
+    session = await builder.connect()
+    result = await session.execute("SELECT * FROM system.local")
+    print(result)
```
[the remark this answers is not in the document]
jakub-graczyk: Tests session creation doesn't brake with execution profile specified.

Lorak-mmk:
@jakub-graczyk I think we can merge that, and add examples in other PR. I think examples were even stated as separate task during our meeting, right?

PR #31, opened 2025-12-28
Add unprepared statement

jakub-graczyk:
This PR implements `Statement` which represents unprepared statement.

This allows users to create not yet prepared statements with specified execution profile or individual options.


src/session.rs
```diff
@@ -53,15 +73,18 @@ impl Session {
...
+            let scylla_statement = statement.extract::<Py<Statement>>(py)?;
+            Ok::<Py<Statement>, PyErr>(scylla_statement)
+        }) {
+            let scylla_statement = statement.get()._inner.clone();
+            return self.scylla_prepare(scylla_statement).await;
         }
+        if let Ok(text) = Python::attach(|py| statement.extract::<String>(py)) {
+            return self.scylla_prepare(text).await;
+        }
+        Err(PyErr::new::<PyTypeError, _>("Invalid statement type"))
```
wprzytula: :wrench: When preparing a `Statement`, the created `PreparedStatement` should inherit the whole configuration from the `Statement`. Please fix the code accordingly.

src/statement.rs
```diff
@@ -98,8 +99,111 @@ impl PreparedStatement {
...
+    #[new]
+    fn new(query_str: String) -> PyResult<Statement> {
+        let s = unprepared::Statement::from(query_str);
+        Ok(Statement { _inner: s })
+    }
+
+    #[getter]
+    fn contents(&self) -> String {
+        self._inner.contents.clone()
+    }
```
wprzytula: :question: Is it needed to clone and allocate the String on the heap? After all, PyO3 converts the resulting String into Python string anyway.

src/statement.rs
```diff
@@ -98,8 +99,111 @@ impl PreparedStatement {
...
+        let mut s = self._inner.clone();
+        s.set_execution_profile_handle(Some(profile._inner.into_handle()));
+        Statement { _inner: s }
+    }
+
+    fn without_execution_profile(&self) -> Statement {
+        let mut s = self._inner.clone();
+        s.set_execution_profile_handle(None);
+        Statement { _inner: s }
+    }
```
wprzytula: :cry: It's very sad to me that we clone the whole Rust `Statement`, including the query string, on any setter.

src/statement.rs
```diff
@@ -98,8 +99,111 @@ impl PreparedStatement {
...
+        s.unset_serial_consistency();
+        Statement { _inner: s }
+    }
+
+    fn get_serial_consistency(&self) -> Option<SerialConsistency> {
+        // TODO: implement returning Unset like in get_request_timeout
+        self._inner
+            .get_serial_consistency()
+            .map(SerialConsistency::to_python)
+    }
```
wprzytula: :question: When is the TODO going to be resolved? What are the difficulties?

Lorak-mmk: My guess is that it will be fixed when we fix Rust Driver to not flatten nested Options in serial consistency getters.

Issue #36, opened 2026-01-02
Define Python driver behavior for non-rows query results


patrycja-ziemkiewicz:
Discussion in deserialization PR: https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/pull/25#discussion_r2658272977

Issue #9, opened 2025-09-15
Implement paging


Lorak-mmk:

wprzytula:
Node.js-RS Driver paging examples: https://github.com/scylladb/nodejs-rs-driver/tree/main/examples/paging

Rust Driver's refactor that hardened paging API to prevent user pitfalls: https://github.com/scylladb/scylla-rust-driver/pull/1061

PR #38, opened 2026-01-11
Paging

patrycja-ziemkiewicz:
This PR adds:

- a paging API for query execution,
- support for non-row query results,
- an async iterator for paged results,
- helper methods first_row() and all() for convenient consumption of paged results,
- additional utilities for working with PagingState,
- tests covering all new added features,
- examples demonstrating interactions with RequestResult.

Fixes: https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/36 , https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/9

### DIFF
--- a/examples/paging.py
+++ b/examples/paging.py
@@ -1,0 +1,261 @@
+"""
+paging.py
+
+Example showcasing multiple ways to consume ScyllaDB query results with the new Python driver API.
+
+This file demonstrates:
+  1) Simple async iteration over all rows (auto-paging under the hood)
+  2) Manual paging: iter_current_page() + fetch_next_page()
+  3) Manual paging with explicit PagingState resume
+  4) Convenience helpers: first_row() and all()
+  5) Custom row shaping via RowFactory
+
+"""
+
+import asyncio
+import os
+from typing import Any, Dict
+
+from scylla.session_builder import SessionBuilder
+from scylla.session import Session
+from scylla.statement import Statement
+from scylla.results import RowFactory, ColumnIterator
+
+
+# ----------------------------
+# DB setup helpers
+# ----------------------------
+async def setup_schema(session: Session) -> None:
+    # Create keyspace & table.
+    await session.execute(
+        """
+        CREATE KEYSPACE IF NOT EXISTS examples_ks
+        WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1};
+        """
+    )
+    await session.execute("USE examples_ks")
+
+    await session.execute(
+        """
+        CREATE TABLE IF NOT EXISTS select_paging (
+                                                     a int,
+                                                     b int,
+                                                     c text,
+                                                     PRIMARY KEY (a, b)
+            );
+        """
+    )
+
+    # Insert a small deterministic dataset.
+    # (Re-inserting is fine: primary key makes rows idempotent for the same keys.)
+    for i in range(16):
+        await session.execute(
+            "INSERT INTO select_paging (a, b, c) VALUES (?, ?, 'abc')",
+            (i, 2 * i),
+        )
+
+
+# ----------------------------
+# 1) Easiest: async for (auto-paging)
+# ----------------------------
+async def example_async_for(session: Session) -> None:
+    print("\n=== 1) Async iteration over all rows (auto-paging) ===")
+
+    # Unprepared string query (supports str | Statement | PreparedStatement).
+    result = await session.execute("SELECT a, b, c FROM select_paging")
+
+    async for row in result:
+        # Default row representation: dict[str, CqlValue]
+        print(f"row={row}")
+
+
+# ----------------------------
+# 2) Manual paging loop: iter_current_page() + fetch_next_page()
+# ----------------------------
+async def example_manual_paging_unprepared(session: Session) -> None:
+    print("\n=== 2) Manual paging (unprepared Statement) ===")
+
+    stmt = Statement("SELECT a, b, c FROM select_paging").with_page_size(6)
+    result = await session.execute(stmt)
+
+    page_no = 1
+    while True:
+        # Consume only the *current* page:
+        page_rows: list[Any] = []
+
+        for res in result.iter_current_page():
+            page_rows.append(res)
+
+        print(f"page {page_no}: {len(page_rows)} rows -> {page_rows}")
+
+        if not result.has_more_pages():
+            break
+
+        # Fetch the next page (returns a new RequestResult):
+        next_result = await result.fetch_next_page()
+
+        # `fetch_next_page()` returning None is an alternative way of detecting
+        # that there are no more pages. In this example we already checked
+        # `has_more_pages()`, so None here would indicate an inconsistent state.
+        assert next_result is not None
+
+        result = next_result
+        page_no += 1
+
+
+async def example_manual_paging_prepared(session: Session) -> None:
+    print("\n=== 3) Manual paging (prepared statement) ===")
+
+    prepared = await session.prepare("SELECT a, b, c FROM select_paging")
+    # Setting page size on the prepared statement applies to all executions of it
+    prepared = prepared.with_page_size(7)
+
+    result = await session.execute(prepared)
+
+    page_no = 1
+    while True:
+        # Consume only the *current* page of size 7 except maybe the last one:
+        page_rows = list(result.iter_current_page())
+        print(f"page {page_no}: {len(page_rows)} rows")
+
+        if not result.has_more_pages():
+            break
+
+        # Fetch the next page (returns a new RequestResult):
+        next_result = await result.fetch_next_page()
+
+        # `fetch_next_page()` returning None is an alternative way of detecting
+        # that there are no more pages. In this example we already checked
+        # `has_more_pages()`, so None here would indicate an inconsistent state.
+        assert next_result is not None
+
+        result = next_result
+        page_no += 1
+
+
+# ----------------------------
+# 3) PagingState: resume later
+# ----------------------------
+async def example_paging_state_resume(session: Session) -> None:
+    print("\n=== 4) PagingState resume ===")
+
+    prepared = await session.prepare("SELECT a, b, c FROM select_paging")
+    prepared = prepared.with_page_size(5)
+
+    # Fetch first page
+    result = await session.execute(prepared)
+
+    seen_rows: list[Any] = []
+
+    while True:
+        page = list(result.iter_current_page())
+        print(f"page size={len(page)}")
+        seen_rows.extend(row for row in page)
+
+        state = result.paging_state()
+
+        # Check if more pages are available via paging state. If None, no more pages.
+        if state is None:
+            break
+
+        # Resume: new execute call with the returned paging_state starts from "after the first page"
+        result = await session.execute(
+            prepared,
+            paging_state=state,
+        )
+
+
+# ----------------------------
+# 4) Convenience helpers: first_row() and all()
+# ----------------------------
+async def example_first_row_and_all(session: Session) -> None:
+    print("\n=== 5) Convenience helpers: first_row() and all() ===")
+
+    prepared = await session.prepare("SELECT a, b, c FROM select_paging")
+    prepared = prepared.with_page_size(4)
+
+    result = await session.execute(prepared)
+
+    # first_row(): returns one row or None (does not force consuming the full result set)
+    one = await result.first_row()
+    print(f"first_row() -> {one}")
+
+    # all(): eagerly fetches all remaining pages and materializes into a list
+    rows = await result.all()
+    print(f"all() -> {len(rows)} rows")
+
+
+# ----------------------------
+# A custom RowFactory example
+# ----------------------------
+class SelectedColumnsDictFactory(RowFactory):
+    """
+    Keep only selected columns in the produced row dict.
+    """
+
+    def __init__(self, columns: list[str]) -> None:
+        super().__init__()
+        self.columns = set(columns)
+
+    def build(self, column_iterator: ColumnIterator) -> Dict[str, Any]:
+        return {col.column_name: col.value for col in column_iterator if col.column_name in self.columns}
+
+
+class UppercaseKeysDictFactory(RowFactory):
+    """
+    Example: dict row, but keys uppercased.
+    """
+
+    def build(self, column_iterator: ColumnIterator) -> Dict[str, Any]:
+        return {col.column_name.upper(): col.value for col in column_iterator}
+
+
+# ----------------------------
+# 5) Custom row shapes (RowFactory)
+# ----------------------------
+async def example_custom_row_factory(session: Session) -> None:
+    print("\n=== 5) Custom row factories ===")
+
+    stmt = Statement("SELECT a, b, c FROM select_paging").with_page_size(20)
+    result = await session.execute(stmt, factory=UppercaseKeysDictFactory())
+
+    res = await result.all()
+    print(f"set_factory(UppercaseKeysDictFactory()); first row -> {res[:1]}")
+
+    # And async iteration will now yield only the selected columns as dict keys:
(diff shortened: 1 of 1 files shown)


python/scylla/_rust/results.pyi
```diff
@@ -144,3 +156,36 @@ class RequestResult:
         are constructed.
         """
         ...
+
+class PagingRequestResult:
```
wprzytula: :thinking: The name is not quite intuitive. Maybe `MultiPageRequestResult`?

src/session.rs
```diff
@@ -117,6 +150,19 @@ impl Session {
...
+        paging_state: PagingState,
+        prepared: scylla::statement::prepared::PreparedStatement,
+    ) -> Result<(QueryResult, PagingStateResponse), PyErr> {
+        self.session_spawn_on_runtime(async move |s| {
+            s.execute_single_page(&prepared, &[], paging_state)
+                .await
+                .map_err(|e| PyRuntimeError::new_err(e.to_string()))
+        })
+        .await
+    }
```
wprzytula: :wrench: this function seems to only need `&PreparedStatement` (not ownership).

src/session.rs
```diff
@@ -1,90 +1,123 @@
...
-        Err(PyErr::new::<PyTypeError, _>("Invalid statement type"))
+    }
+}
+
+#[derive(Clone)]
+pub(crate) enum QueryRequest {
+    Prepared(statement::prepared::PreparedStatement),
+    Text(String),
+    Statement(unprepared::Statement),
+}
```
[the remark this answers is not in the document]
wprzytula:
```suggestion
#[derive(Clone)]
pub(crate) enum ExecutableStatement {
    Prepared(statement::prepared::PreparedStatement),
    Text(String),
    Statement(unprepared::Statement),
}
```

src/session.rs
```diff
@@ -1,90 +1,123 @@
...
-            return self.scylla_prepare(scylla_statement).await;
-        }
-        if let Ok(text) = Python::attach(|py| statement.extract::<String>(py)) {
-            return self.scylla_prepare(text).await;
+        let query_request = QueryRequest::extract_request(statement)?;
+
+        match query_request {
+            QueryRequest::Statement(s) => self.scylla_prepare(s).await,
+            QueryRequest::Text(sql) => self.scylla_prepare(sql).await,
+            _ => Err(PyErr::new::<PyTypeError, _>("Invalid statement type")),
```
wprzytula: :wrench: Insufficient information in the error message. What statement type was received and what was expected?

src/session.rs
```diff
@@ -1,90 +1,123 @@
...
+    #[pyo3(signature = (request, paging_state=None, page_size=None))]
+    async fn execute_paged(
+        &self,
+        request: Py<PyAny>,
+        paging_state: Option<Py<PyPagingState>>,
+        page_size: Option<i32>,
+    ) -> PyResult<PagingRequestResult> {
+        let query_request = QueryRequest::extract_request(request)?;
+
+        //Ensure the query is prepared so it can be efficiently reused while fetching pages.
```
wprzytula: :pick: Please put space after `//`.

python/scylla/_rust/statement.pyi
```diff
@@ -22,6 +22,8 @@ class PreparedStatement:
     def with_request_timeout(self, timeout: float | None) -> PreparedStatement: ...
     def without_request_timeout(self) -> PreparedStatement: ...
     def get_request_timeout(self) -> float | None | UnsetType: ...
+    def with_page_size(self, page_size: int) -> PreparedStatement: ...
+    def get_page_size(self) -> int: ...
```
Lorak-mmk: Introducing those configurations on statement could be done in a separate commit, in order to simplify the "main" commit of this series.

python/scylla/_rust/results.pyi
```diff
@@ -194,6 +194,8 @@ class RequestResult:
         ...
 
     def __aiter__(self) -> AsyncRowsIterator: ...
+    
+    def __iter__(self) -> SyncRowsIterator: ...
```
Lorak-mmk: This will be too easy to use accidentaly, blocking the reactor. Can we force the user to call a method like `blocking_paging_iter`?

src/session.rs
```diff
@@ -32,9 +49,40 @@ impl Session {
...
+    #[pyo3(signature = (request, paging_state=None, page_size=None))]
+    async fn execute_paged(
+        &self,
+        request: Py<PyAny>,
+        paging_state: Option<Py<PyPagingState>>,
+        page_size: Option<i32>,
+    ) -> PyResult<RequestResult> {
+        let query_request = ExecutableStatement::extract_request(request)?;
+
+        //Ensure the query is prepared so it can be efficiently reused while fetching pages.
```
wprzytula: :thinking: I'm not sure we should do such thing. If user wanted to have statements prepared, they would do that beforehand.

examples/paging.py
```diff
@@ -0,0 +1,264 @@
...
+        )
+
+
+# ----------------------------
+# 1) Easiest: async for (auto-paging)
+# ----------------------------
+async def example_async_for(session: Session) -> None:
+    print("\n=== 1) Async iteration over all rows (auto-paging) ===")
+
+    # Unprepared string query (your API supports str | Statement | PreparedStatement).
```
wprzytula: 🤔 Who's _you_ in `your API`? I suppose it's the person the LLM is talking to.

examples/paging.py
```diff
@@ -0,0 +1,264 @@
...
+    print("\n=== 5) Convenience helpers: first() and all() ===")
+
+    prepared = await session.prepare("SELECT a, b, c FROM select_paging")
+    prepared = prepared.with_page_size(4)
+
+    result = await session.execute(prepared)
+
+    # first(): returns one row or None (does not force consuming the full result set)
+    one = await result.first()
+    print(f"first() -> {one}")
```
wprzytula: ❓ What is the state of the result set once we call `first()`? Can I call `first()` repeatedly? Is that equivalent to `__next__()`?
If they are equivalent, what's the use of `first` if one can just call `next()` builtin function?

python/tests/test_deserialization.py
```diff
@@ -70,7 +70,7 @@ async def insert_and_fetch_single_row(
     await session.execute(f"INSERT INTO {table} (id, value) VALUES ({row_id}, {value_sql});")
 
     result = await session.execute(f"SELECT * FROM {table} WHERE id = {row_id}")
-    return next(result.iter_rows())
+    return next(result.iter_page())
```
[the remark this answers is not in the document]
wprzytula: I still believe this is not a serious problem with the API. Let's warn in `iter_current_page`'s docstring that the first page may as well be empty. Also, `iter_current_page` is not expected to be the main API - `async for` is.

src/deserialize/results.rs
```diff
@@ -69,34 +150,213 @@ impl RequestResult {
...
+    fn paging_state(&self) -> Option<PyPagingState> {
+        match self {
+            QueryPager::Paged {
+                paging_response: PagingStateResponse::HasMorePages { state },
+                ..
+            } => Some(PyPagingState {
+                inner: state.clone(),
+            }),
+            _ => None,
+        }
```
wprzytula: ⛏️ Let's avoid the wildcard if possible. `QueryPager::Unpaged` is just one variant to be listed there.

src/deserialize/results.rs
```diff
@@ -69,34 +150,213 @@ impl RequestResult {
...
+            value_list,
+        } = self
+        else {
+            return None;
+        };
+
+        let state = match paging_response {
+            PagingStateResponse::HasMorePages { state } => state.clone(),
+            _ => return None,
+        };
```
wprzytula: ⛏️ Either use `let-else` or replace the wildcard with the explicit pattern.

src/deserialize/results.rs
```diff
@@ -69,34 +150,213 @@ impl RequestResult {
...
+
+        let result = session
+            .execute_single_page(state, query_request.clone(), value_list.clone())
+            .await;
```
wprzytula: 💭 @Lorak-mmk Here we could benefit from BoundStatement once it's introduced, because serialization would only happen once.

src/deserialize/results.rs
```diff
@@ -69,34 +150,213 @@ impl RequestResult {
...
+
+        pyo3_async_runtimes::tokio::future_into_py(py, async move {
+            let mut rows_iterator = rows_iterator_clone.lock().await;
+            let mut query_pager = query_pager_clone.lock().await;
+
+            next_row_with_paging(rows_iterator.deref_mut(), query_pager.deref_mut())
+                .await
+                .unwrap_or(Err(PyErr::new::<PyStopAsyncIteration, _>("")))
+        })
+    }
```
wprzytula: ❓ Will you run such benchmarks? Let's start with something simple.

examples/paging.py
```diff
@@ -0,0 +1,264 @@
...
+# ----------------------------
+async def example_custom_row_factory(session: Session) -> None:
+    print("\n=== 7) Custom row factories ===")
+
+    stmt = Statement("SELECT a, b, c FROM select_paging").with_page_size(5)
+    result = await session.execute(stmt)
+
+    # Option A: per-iterator factory (does not change defaults)
+    tuples = list(result.iter_page(factory=TupleRowFactory()))
+    print(f"iter_page(factory=TupleRowFactory()) -> {tuples[:2]} ...")
```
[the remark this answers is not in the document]
patrycja-ziemkiewicz: > 💭 Is concurrent iteration over the result set supported? What if two threads concurrently call `__anext__()` on the result set? What if one thread alternates between `__anext__()` on the results set and `__next__()` on a page iterator obtained previously using `iter_page()`?

`__anext__` iteration is thread safe. 
`iter_page()` is not (probably should be changed in future).

The scenario described above works correctly because each of __aiter__(), iter_page(), and blocking_paging_iter() creates an independent iterator. Advancing one iterator does not affect the
state of any other iterator.

In other words, the same result set can currently be iterated over multiple times without exhausting it. We should probably discuss if this is the desired approach and if not I will change it. And regarding of the outcome I should definitely describe it precisely in documentation.

examples/paging.py
```diff
@@ -0,0 +1,264 @@
...
+    def consume_blocking() -> int:
+        # This blocks the current thread while pages are fetched.
+        count = 0
+        for row in result.blocking_paging_iter():
+            count += 1
+            print(f"row={row}")
+        return count
+
+    total = await asyncio.to_thread(consume_blocking)
+    print(f"blocking_paging_iter() consumed {total} rows")
```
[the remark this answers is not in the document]
patrycja-ziemkiewicz: This blocking_paging_iter() was introduce mostly for benchmarking. It’s likely not something we plan to expose in the new driver API. To be consistent We would require a truly blocking execute(), which is not planned for the new driver, but which would fit in a legacy API.

examples/paging.py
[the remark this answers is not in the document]
patrycja-ziemkiewicz: When I thought about it longer it probably would be more confusing than beneficial to have non exhausting result I will just change it.

examples/paging.py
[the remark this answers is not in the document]
wprzytula: Oh really? So we want to have async iteration exclusively, right? (and current page iteration apart from that).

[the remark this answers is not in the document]
Lorak-mmk: It would be fine to have a separate draft PR with them and decide their fate there

src/session.rs
```diff
@@ -1,50 +1,49 @@
...
 impl Session {
-    #[pyo3(signature = (statement, values = None))]
+    #[pyo3(signature = (statement, values=None, factory=None, paging_state=None, paged=true))]
     async fn execute(
         &self,
         statement: ExecutableStatement,
         values: Option<PyValueList>,
+        factory: Option<Py<RowFactory>>,
+        paging_state: Option<Py<PyPagingState>>,
+        paged: bool,
```
Lorak-mmk: We should think about this signature a bit. Specifically about which arguments should be positional only, which should be keyword only, and which should be both.
`statement` and `values` should imo be positional only, as they are passed basically always, and their order and meaning is intuitive.
The rest could be made keyword only, so that their meaning at the callsites is clear.

In that case the signature would be `statement, values=None, /, *, factory=None, paging_state=None, paged=true`

python/tests/test_prepare.py
```diff
@@ -0,0 +1,52 @@
...
+@pytest.mark.requires_db
+async def test_prepare_and_str():
+    builder = SessionBuilder(["127.0.0.2"], 9042)
+    session = await builder.connect()
+    query_str = "SELECT cluster_name FROM system.local"
+    prepared = await session.prepare(query_str)
+    result_prepared = await session.execute(prepared)
+    result_str = await session.execute(query_str)
+    assert list(result_prepared.iter_current_page()) == list(result_str.iter_current_page())
+
```
Lorak-mmk: If user wants to use paging (which should be the default), then driver should scream loudly if not all pages were exhausted. What you wrote in this test is exactly how paging should not be used.
Please either use unpaged here, or at least check there are no more pages.

And as I said, we should have a guardrail that throws (or at least logs) if there were unused pages.

python/scylla/_rust/results.pyi
```diff
@@ -183,6 +183,30 @@ class RequestResult:
...
+        Fetches the first available row from the current page onwards,
+        automatically retrieving additional pages as needed.
+        Returns None if no more rows are available.
+
+        Returns
+        -------
+        Any | None
+            The first row as a Python object, or None if no more rows exist.
+        """
+        ...
```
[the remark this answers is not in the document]
wprzytula: Let's focus on saying in the first place that `RequestResult` is immutable. Many conclusions come from this fact.

python/scylla/_rust/session.pyi
```diff
@@ -1,12 +1,57 @@
+from .results import PagingState, RequestResult, RowFactory
 from typing import Any
 
-from .results import RequestResult
 from .statement import PreparedStatement, Statement
 
 class Session:
+    """
+    Represents a CQL session, which can be used to communicate with the database
```
wprzytula:
```suggestion
    Represents a CQL session, which can be used to communicate with the database.
```

src/deserialize/results.rs
```diff
@@ -1,102 +1,381 @@
...
+    rows_iterator: RowsIteratorKind,
+    query_pager: QueryPager,
+}
+
+async fn next_row_with_paging(
+    rows_iterator: &mut RowsIteratorKind,
+    query_pager: &mut QueryPager,
+) -> Option<PyResult<Py<PyAny>>> {
+    // Loop until a row is produced, all pages are exhausted,
+    // or an error occurs while fetching or updating pages.
```
wprzytula:
```suggestion
/// Loop until a row is produced, all pages are exhausted,
/// or an error occurs while fetching or updating pages.
async fn next_row_with_paging(
    rows_iterator: &mut RowsIteratorKind,
    query_pager: &mut QueryPager,
) -> Option<PyResult<Py<PyAny>>> {
```

src/deserialize/results.rs
```diff
@@ -1,102 +1,381 @@
...
+        let query_result = match query_pager.fetch_next_page().await? {
+            Ok(p) => p,
+            Err(e) => return Some(Err(e)),
+        };
+
+        if let Err(err) = Python::attach(|py| rows_iterator.update(py, Arc::new(query_result))) {
+            return Some(Err(err));
+        }
+    }
+}
```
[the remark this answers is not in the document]
Lorak-mmk: The differences in body of this function in both variants are minimal.
The current variant (`Option<Result<...>>`) makes more sense to me (iterator of results, like in TypedRowIterator), and matches other parts of the code (`fetch_next_page`, `rows_iterator.next`).
So I vote for keeping the current variant.

PR #40, opened 2026-01-19
Feature/error handling

### LINKED ISSUE #28: Error handling
Currently all error handling in the driver is done in quite ad-hoc way. We have to choose an approach to take, and abide by it.
Before that, we need to understand well what is the approach to errors in pyo3 - how to create them, transform them etc. Please research that and write down conclusions.

The big pain point in old Python Driver (and in many Python libraries tbh) is that you never know what errors can be thrown from any function, and when.
Because Rust Driver has a nice strongly typed error hierarchy, we have the opportunity to improve that in our driver.

With that in mind, design and propose an approach to error handling: 
 - how to translate Rust Driver errors to Python
 - types used for our errors
 - how to document error conditions
 - how to achieve as much compile-time guarantees as possible (so less `PyErr` and more strong typing is better).

pasinskik:
## Pre-review checklist

- [X] I added appropriate `Fixes:` annotations to PR description.

## Summary

This PR introduces a dedicated error handling layer for the Python driver bindings.

The main goal is to provide Python users with clear, structured, and idiomatic exceptions, while keeping the internal Rust-side error representation explicit and manageable. Instead of leaking raw Rust driver errors or generic Python exceptions at arbitrary points, errors are now modeled in Rust and converted to appropriate Python exception classes at the API boundary.

## What changed

This PR adds:

- a hierarchy of custom Python exception classes for the bindings
- Rust-side error types covering the main failure categories exposed by the driver
- `From<...> for PyErr` conversions that map internal driver errors to user-facing Python exceptions
- improved error messages with additional context where available
- preservation of original Python exceptions as causes when the failure originates from Python code
- structured error attributes for easier inspection from Python code

In particular:

- deserialization errors can carry structured location information such as column name, column index, and nested path
- serialization errors can carry parameter reference information
- session/configuration/statement/batch/execute/prepare/schema-agreement failures are converted into dedicated exception types instead of generic runtime errors

## Why

The previous approach relied more heavily on generic Python exceptions and ad hoc conversion, which made it harder to:

- distinguish different classes of failures from Python
- preserve the original source of an error
- provide consistent, user-friendly messages
- attach structured debugging information in a predictable way

This change makes error handling more explicit and more useful for both users and maintainers.

For Python users, this means:

- catching more specific exceptions
- better error messages
- access to additional context attached as exception attributes
- preserved exception chaining for failures originating in Python callbacks/conversions

For the implementation, this means:

- clearer separation between internal Rust-side error modeling and Python-facing exception conversion
- a single, consistent conversion point at the boundary
- room to extend the error hierarchy in a structured way as the API grows

## Design notes

The policy in this PR is:

- internal Rust-side error types represent failures in a strongly typed way
- conversion to `PyErr` happens at the Python boundary
- when the error originates from Python code, the original `PyErr` is preserved as the cause
- when the error originates from the Rust driver or binding layer, a dedicated Python exception is created with a descriptive message and any relevant structured attributes

This keeps the Rust side ergonomic and typed, while making the Python side idiomatic and debuggable.

Fixes: #28


Lorak-mmk (changes requested):
Review of first commit: "errors: add structured driver errors for deserialization"

src/errors.rs
```diff
@@ -0,0 +1,370 @@
...
+// using the From<DriverError> for PyErr implementation, which maps each DriverError variant to
+// an appropriate Python exception class and attaches any relevant information as attributes or causes.
+
+/* Rust errors */
+#[derive(Debug)]
+#[allow(dead_code)] // we will have more variants here in the future
+pub(crate) enum DriverError {
+    Deserialization(DriverDeserializationError),
+}
+
```
Lorak-mmk: Please don't use `allow`, use `expect` instead. That way you won't forget it when it actually starts to be used.

src/errors.rs
```diff
@@ -336,16 +359,210 @@ impl From<DriverDeserializationError> for PyErr {
...
 }
 
+/* Execution errors */
+
+/// Errors that can occur during execution of queries, connection establishment, or session configuration.
+#[derive(Debug)]
+pub enum DriverExecutionError {
+    ConnectionError(ConnectionError),
+    SessionConfigError(SessionConfigError),
+}
```
Lorak-mmk: Its the same here. What do we need that for? Certainly not `execute` method, because `ConnectionError` and `SessionConfigError` don't make any sense there.

Lorak-mmk (changes requested):
Quite a lot of comments. Many of them related to the commit structure: it seems there are a lot of places where some change is in a wrong commit, or fixes an issue introduced in earlier commit.

src/errors.rs
```diff
@@ -0,0 +1,851 @@
...
+    /// Driver invariant violated: a deserializer was called for a mismatched ColumnType.
+    /// This indicates a bug in our dispatch logic.
+    WrongDeserializer { expected: &'static str, got: String },
+}
+
+impl DriverDeserializationError {
+    /* Constructors */
+
+    #[must_use]
+    pub fn unsupported_type(cql: impl Into<String>) -> Self {
```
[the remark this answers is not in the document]
wprzytula: This should have the same effect, so let's do so.

src/deserialize/value.rs
```diff
@@ -596,63 +657,92 @@ fn deser_cql_py_value<'py, 'metadata, 'frame>(
...
                     }
                     // CQL Date → Python datetime.date
                     NativeType::Date => {
-                        let date: chrono::NaiveDate = CqlDate::deserialize(typ, Some(v))?
+                        let date: chrono::NaiveDate = CqlDate::deserialize(typ, Some(v))
+                            .map_err(DriverDeserializationError::scylla_decode_failed)?
                             .try_into()
-                            .map_err(DeserializationError::new)?;
+                            .map_err(DeserializationError::new)
+                            .map_err(DriverDeserializationError::scylla_decode_failed)?;
```
patrycja-ziemkiewicz: I noticed this pair of map_err calls is repeated in several places. It might be worth extracting it into a small helper function to reduce duplication and improve readability.

Lorak-mmk (changes requested):
Didn't review serialization / deserialization yet.

src/errors.rs
```diff
@@ -197,10 +201,138 @@ impl From<SessionConfigError> for PyErr {
...
+    /// paging_state parameter in session.execute must be None.
+    PagingStateMustBeNoneForUnpagedExecution,
+    /// The Rust driver failed while executing a query.
+    StatementExecutionError {
+        source: Box<scylla::errors::ExecutionError>,
+    },
+    /// The Rust driver failed while preparing a statement.
+    StatementPrepareError {
+        source: Box<scylla::errors::PrepareError>,
+    },
```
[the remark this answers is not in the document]
pasinskik: I have adjusted the errors, changed `String` to `Box<str>` in places where it was crucial, I also switched `Vec<InnerSegment>` for `Box<[InnerSegment]>`.

Current error sizes are as follows:

|Error Name                |Size|
|--------------------------|----|
|RowIterationError         |72 B|
|DriverDeserializationError|72 B|
|ConnectionError           |24 B|
|SessionConfigError        |24 B|
|StatementConversionError  |24 B|
|ExecuteError              |24 B|
|PrepareError              |8 B |
|SchemaAgreementError      |16 B|
|StatementConfigError      |16 B|
|BatchError                |16 B|
|DriverSerializationError  |48 B|

src/errors.rs
```diff
@@ -0,0 +1,31 @@
...
+// and attach any relevant structured information as attributes.
+
+// For errors originating from our own Rust code, we create a custom Python exception with a descriptive message,
+// and we can include any relevant information in the message or as attributes.
+
+#[pymodule]
+pub(crate) fn errors(py: Python<'_>, module: &Bound<'_, PyModule>) -> PyResult<()> {
+    module.add("ScyllaError", py.get_type::<ScyllaErrorPy>())?;
+    Ok(())
+}
```
Lorak-mmk, line 1257: I assume this will be applicable to all exception types, so I'll write this once:
We want users to not see the `Py` suffixes. In other words, user should only ever see `ScyllaError` name, not `ScyllaErrorPy` name. Is that the case now? I don't think so.  I looked at what the macro expands to, and there is a problematic section:
```rust
unsafe impl pyo3::type_object::PyTypeInfo for ScyllaErrorPy {
    const NAME: &'static str = "ScyllaErrorPy";
    const MODULE: ::std::option::Option<&'static str> = ::std::option::Option::Some("errors");
    #[inline]
    #[allow(clippy::redundant_closure_call)]
    fn type_object_raw(py: pyo3::Python<'_>) -> *mut pyo3::ffi::PyTypeObject {
        use pyo3::sync::PyOnceLock;
        static TYPE_OBJECT: PyOnceLock<pyo3::Py<pyo3::types::PyType>> = PyOnceLock::new();
        TYPE_OBJECT
            .get_or_init(py, || {
                pyo3::PyErr::new_type(
                    py,
                    pyo3_ffi::_cstr_from_utf8_with_nul_checked("errors.ScyllaErrorPy\u{0}"),
                    (::std::option::Option::None),
                    ::std::option::Option::Some(&py.get_type::<PyException>()),
                    ::std::option::Option::None,
                )
                .expect("Failed to initialize new exception type.")
            })
            .as_ptr()
            .cast()
    }
}
```

I suspect the names passed to `create_exception!` will need to be changed to not include `Py` suffix.

python/scylla/_rust/errors.pyi
```diff
@@ -1,3 +1,5 @@
 from __future__ import annotations
 
 class ScyllaError(Exception): ...
+class ConnectionError(ScyllaError): ...
+class SessionConfigError(ScyllaError): ...
```
Lorak-mmk, line 15: Ah, so `ConnectionError` is an error that happened when setting up session.
Can we name it differently? For developers / users of Rust Driver it will be confusing, because there is `ConnectionError` which is an error of a single connection (translation error, broken connection etc).

Lorak-mmk (changes requested):
I think we need to merge this PR quickly, and it should be our top priority for now.

I realise the blame is mostly on us for not doing review quickly enough, sorry for that.
Starting now, I won't be waiting with posting comment until I finish full review, I'll just post what I spot immediately. This should improve our iteration time.

Lorak-mmk (commented):
I looked at all the commits, and I think we can merge after you fix the 2 comments above.

I only did a very brief reading, but I believe we should merge it quickly, to avoid endless conflicts. We can improve later if needed.

python/scylla/_rust/errors.pyi
wprzytula: `SessionConnectionError`?
`ClusterConnectionError`?
`SessionStartupError`?

Lorak-mmk: All are fine with me, feel free to choose.

src/errors.rs
pasinskik: I investigated this and I think you're right.

I need to change the exceptions' names, but then also Rust error enums need to be renamed to avoid name collisions.

I think we could add `Driver` prefix to internal Rust errors, i.e. `DriverPrepareError`. I have already done this with `DriverDeserializationError` and `DriverSerializationError`, so it would make the file more consistent too.

python/scylla/_rust/errors.pyi
pasinskik: I will rename it to `SessionConnectionError` for clarification.

src/errors.rs
Lorak-mmk: "Driver" prefix should be fine, "Internal" could also fit. @wprzytula ?

pasinskik: I would prefer `Driver` here, to me `Internal` suggest a bug in the driver's logic or something that a user could not reasonably trigger except by hitting a defect. That's why the messages in `RuntimeTaskJoinFailed` and `WrongDeserializer` exception instances start with "internal error:"

src/deserialize/results.rs
```diff
@@ -527,22 +551,22 @@ impl RowFactory {
     ///
     /// Raises
     /// ------
-    /// RuntimeError
-    ///     If deserialization of any column fails.
+    /// DeserializationError
+    ///     If any column cannot be deserialized into a Python object.
+    /// Exception
+    ///     If building the Python row object fails.
```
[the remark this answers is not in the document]
pasinskik: Fixed. Actually, this function can raise `DeserializationError` (because `RowIterationError::DeserializationError` gets converted to a `DeserializationError` exception). At other times `RowIterationError` is raised (from `RowIterationError::PythonError`) with the original PyErr attached as cause.

PR #45, opened 2026-02-19
Refractoring

patrycja-ziemkiewicz:
Changes

1. Added docstrings for methods and classes in results.pyi
2. Refactored session-related Rust (session.rs) methods for improved structure and clarity
3. Added setters for configuring page size on:
    - prepared statements
    - statements

## Pre-review checklist


src/session.rs
```diff
@@ -53,84 +55,44 @@ fn is_empty_row(row: &Bound<'_, PyAny>) -> bool {
...
-                .await?; // Propagate error form closure
-            return Ok(RequestResult {
-                inner: Arc::new(result),
-            });
-        }
-        Err(PyErr::new::<PyTypeError, _>("Invalid request type"))
+        let query_request = ExecutableStatement::extract_request(request)?;
+        let result = self
+            .session_spawn_on_runtime(async move |s| {
+                match query_request {
```
wprzytula: ⛏️ Naming: _request_ is not the same as _statement_. _statement_ is the passive object representing the statement to be executed against the DB. _request_ is a single execution of a statement against the DB (in case of `EXECUTE` or `QUERY` CQL request).

Please use `statement` instead of `request` everywhere here.

PR #46, opened 2026-02-19
Add load balancing policy

jakub-graczyk:
### Overview
This PR adds ability to **customize default LBP** as well as **user-defined LBP** in python.

For that purpose exposes metadata about cluster which is needed when defining `fallback()`.

In addition exposes methods on execution profile and statements to use LBP while executing queries.

Tests for implementation are also included.

### Design decisions
- LoadBalancingPolicy is exposed as a Protocol - python's trait equivalent. Object that correctly implements `fallback()` can be used as LBP.
- Necessary metadata must be owned when in python object which facilitates copying in their respective wrappers.
- Squashing methods (eg. `strategy_name()` on `Keyspace`) aims to reduce allocations using references instead of copying the struct (`Strategy` in this case).
- `Arc<dyn LoadBalancingPolicy>` can't be downcast back to the original python object, `ExecutionProfile`, `Statement`, and `PreparedStatement` all store `Option<PyLoadBalancingPolicy>` alongside the rust inner type so `get_load_balancing_policy()` returns the exact same python object by incrementing reference counter.

### DIFF
--- a/src/policies/load_balancing/mod.rs
+++ b/src/policies/load_balancing/mod.rs
@@ -2,6 +2,7 @@ use std::convert::Infallible;
 use std::sync::{Arc, Mutex};
 
 use pyo3::intern;
+use pyo3::sync::MutexExt;
 use pyo3::types::PyIterator;
 use pyo3::{prelude::*, types::PyTuple};
 use scylla::{
@@ -142,7 +143,6 @@ impl DefaultPolicy {
     }
 }
 
-#[expect(dead_code)]
 pub(crate) struct PyLoadBalancingPolicy {
     pub(crate) _inner: Py<PyAny>,
     pub(crate) cluster_cache: Mutex<Option<(Arc<cluster::ClusterState>, Py<ClusterState>)>>,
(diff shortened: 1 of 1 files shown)


src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,354 @@
...
+    fn views<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyDict>> {
+        let dict = PyDict::new(py);
+        for (name, view) in &self._inner.views {
+            let py_view = Py::new(
+                py,
+                MaterializedView {
+                    _inner: view.clone(),
+                },
+            )?;
+            dict.set_item(PyString::new(py, name), py_view)?;
```
patrycja-ziemkiewicz: Same here

src/policies/load_balancing/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub(crate) _inner: Py<PyAny>,
+}
+
+impl Clone for PyLoadBalancingPolicy {
+    fn clone(&self) -> Self {
+        PyLoadBalancingPolicy {
+            _inner: Python::attach(|py| self._inner.clone_ref(py)),
+        }
+    }
+}
```
jakub-graczyk: You can add derive clone

src/policies/load_balancing/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+            _consistency: routing_info.consistency,
+            _serial_consistency: routing_info.serial_consistency,
+            _token: routing_info.token,
+            _table: routing_info
+                .table
+                .map(|t| (t.ks_name().to_string(), t.table_name().to_string())),
+            _is_confirmed_lwt: routing_info.is_confirmed_lwt,
+        }
+    }
+}
```
Lorak-mmk: Don't use `#[allow(dead_code)]` - use `#[expect(dead_code)]`

python/scylla/_rust/policies/load_balancing.pyi
```diff
@@ -0,0 +1,33 @@
...
+    def is_confirmed_lwt(self) -> bool: ...
+
+class DefaultPolicy:
+    def __init__(
+        self,
+        preferred_datacenter: str | None = None,
+        preferred_datacenter_and_rack: tuple[str, str] | None = None,
+        token_aware: bool | None = None,
+        permit_dc_failover: bool | None = None,
+        enable_shuffling_replicas: bool | None = None,
```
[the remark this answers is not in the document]
jakub-graczyk: Updated

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+
+impl ColumnKind {
+    fn from_rust(kind: &cluster::metadata::ColumnKind) -> Self {
+        match kind {
+            cluster::metadata::ColumnKind::Regular => ColumnKind::Regular,
+            cluster::metadata::ColumnKind::Static => ColumnKind::Static,
+            cluster::metadata::ColumnKind::Clustering => ColumnKind::Clustering,
+            cluster::metadata::ColumnKind::PartitionKey => ColumnKind::PartitionKey,
+            _ => ColumnKind::Regular,
+        }
```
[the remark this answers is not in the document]
jakub-graczyk: This enum is non_exhaustive...

PR #48, opened 2026-02-21
Deserialization fixes

patrycja-ziemkiewicz:
This PR introduces fixes to the yoke deserialization logic and improves driver performance by removing redundant GIL acquisitions. It also reorganize code ordering for better maintainability and clarity.

### DIFF
--- a/src/deserialize/results.rs
+++ b/src/deserialize/results.rs
@@ -97,60 +97,6 @@ impl RowsIterator {
     }
 }
 
-/// Yoke-backed wrapper holding row and column iterators.
-///
-/// `Cursor` is stored inside a `Yoke` so that both the row iterator
-///  and the column iterator can borrow from the same data without cloning.
-///
-/// - `next_row` advances the row iterator and switches the active column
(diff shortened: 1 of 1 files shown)


src/deserialize/results.rs
```diff
@@ -196,17 +191,34 @@ impl RowColumnCursor {
...
+        let raw_col = cursor
+            .current_raw_column
+            .as_ref()
+            .ok_or_else(|| PyErr::new::<PyStopIteration, _>(""))?;
+
+        let value = PyDeserializedValue::deserialize_py(raw_col.spec.typ(), raw_col.slice, py)?;
+        let column_name = PyString::new(py, raw_col.spec.name()).unbind();
+
+        Ok(Column { column_name, value })
+    }
```
wprzytula: 🤔 Why is `_internal` function `pub`?

wprzytula: Also, I don't see a need to add the `_internal` suffix. Let's just call it `next_column`.

PR #55, opened 2026-03-17
Python driver benchmarks

patrycja-ziemkiewicz:
This PR add benchmark scenarios for the python-rs driver

Example plots generated by the benchmarker:
image
image

### DIFF
--- a/benchmark/common.py
+++ b/benchmark/common.py
@@ -1,18 +1,9 @@
-from typing import Callable, Awaitable, Any
-
-from datetime import datetime, time, date, timezone
-
 import ipaddress
 import uuid
+from datetime import date, datetime, time, timezone
+from typing import Any
+
 from dateutil.relativedelta import relativedelta
-from scylla.statement import Statement
-
-from scylla.session import Session
-
-from scylla.session_builder import SessionBuilder
-import os
-import asyncio
-
 
 SIMPLE_INSERT_QUERY = "INSERT INTO benchmarks.basic (id, val) VALUES (?, ?)"
 COMPLEX_INSERT_QUERY = "INSERT INTO benchmarks.complex (id, val, tuuid, ip, date, time, tuple, udt, set1, duration) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)"
@@ -20,75 +11,6 @@ COMPLEX_SELECT_QUERY = "SELECT * FROM benchmarks.complex USING TIMEOUT 120s;"
 COMPLEX_SELECT_COUNT = "SELECT COUNT(*) FROM benchmarks.complex USING TIMEOUT 120s;"
 SIMPLE_SELECT_COUNT = "SELECT COUNT(*) FROM benchmarks.basic USING TIMEOUT 120s;"
 SIMPLE_SELECT_QUERY = "SELECT * FROM benchmarks.basic"
-
-
-async def connect() -> Session:
-    uri = os.getenv("SCYLLA_URI", "127.0.0.2:9042")
-    builder = SessionBuilder().contact_points(uri)
-    return await builder.connect()
-
-
-async def init_common() -> Session:
-    session = await connect()
-
-    await session.execute("""
-       CREATE KEYSPACE IF NOT EXISTS benchmarks WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1 }
-    """)
-
-    await session.execute("USE benchmarks")
-    return session
-
-
-async def init_simple_table() -> Session:
-    session = await init_common()
-    await simple_table_cleanup(session)
-    await session.execute("CREATE TABLE benchmarks.basic (id uuid, val int, PRIMARY KEY(id))")
-    return session
-
-
-async def init_complex_table() -> Session:
-    session = await init_common()
-    await complex_table_cleanup(session)
-    await session.execute("CREATE TYPE benchmarks.udt1 (field1 text, field2 int)")
-    await session.execute(
-        "CREATE TABLE benchmarks.complex (id uuid, val int, tuuid timeuuid, ip inet, date date, time time, tuple frozen<tuple<text, int>>, udt frozen<udt1>, set1 set<int>, duration duration, PRIMARY KEY(id))"
-    )
-    return session
-
-
-async def init_with_inserts(
-    num_of_rows: int,
-    query: str,
-    init: Callable[[], Awaitable[Session]],
-    get_data: Callable[[], Any],
-) -> Session:
-    session = await init()
-
-    statement = Statement(query)
-    prepared = await session.prepare(statement)
-
-    tasks = [session.execute(prepared, get_data(), paged=False) for _ in range(num_of_rows)]
-
-    await asyncio.gather(*tasks)
-
-    return session
-
-
-async def simple_table_cleanup(session: Session):
-    await session.execute("DROP TABLE IF EXISTS benchmarks.basic")
-
-
-async def complex_table_cleanup(session: Session):
-    await session.execute("DROP TABLE IF EXISTS benchmarks.complex")
-    await session.execute("DROP TYPE IF EXISTS benchmarks.udt1")
-
-
-async def check_row_cnt(session: Session, cnt: int, query: str):
-    res = await session.execute(query, None, paged=False)
-
-    count = await res.first_row()
-    assert count is not None
-    assert count["count"] == cnt
 
 
 def get_simple_data() -> tuple[uuid.UUID, int]:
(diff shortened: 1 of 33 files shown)


benchmark/common.py
```diff
@@ -0,0 +1,125 @@
...
+
+    await session.execute("USE benchmarks")
+    await session.execute("DROP TABLE IF EXISTS benchmarks.basic")
+    await session.execute(schema)
+    return session
+
+
+async def init_simple_table() -> Session:
+    return await init_common("CREATE TABLE benchmarks.basic (id uuid, val int, PRIMARY KEY(id))")
+
```
Lorak-mmk: Why table is dropped and recreated, while keyspace uses `IF NOT EXISTS`?

benchmark/common.py
```diff
@@ -0,0 +1,125 @@
...
+from scylla.session_builder import SessionBuilder
+import os
+
+
+SIMPLE_INSERT_QUERY = "INSERT INTO benchmarks.basic (id, val) VALUES (?, ?)"
+COMPLEX_INSERT_QUERY = "INSERT INTO benchmarks.basic (id, val, tuuid, ip, date, time, tuple, udt, set1, duration) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)"
+SELECT_COUNT = "SELECT COUNT(1) FROM benchmarks.basic USING TIMEOUT 120s;"
+SELECT_QUERY = "SELECT * FROM benchmarks.basic"
+
```
Lorak-mmk, line 14: I think `COUNT(*)` is more typically used. Can we switch to it?

benchmark/common.py
Lorak-mmk: Also, why do you drop `benchmarks.basic` but not `benchmarks.udt1`?

Lorak-mmk: hmm... `benchmarks.basic` is used for 2 different table schemas. Do we have to do it this way? Can we use `benchmarks.complex` for the table created in `init_complex_table`?

benchmark/common.py
```diff
@@ -0,0 +1,125 @@
...
+    get_data: Callable[[], Any],
+) -> Session:
+    session = await init()
+
+    statement = Statement(query)
+    prepared = await session.prepare(statement)
+
+    for _ in range(num_of_rows):
+        await session.execute(prepared, get_data(), paged=False)
+
```
Lorak-mmk: Let's do this concurrently, preferably with concurrency limited to some reasonable value.

benchmark/common.py
```diff
@@ -0,0 +1,125 @@
...
+    return await init_common("CREATE TABLE benchmarks.basic (id uuid, val int, PRIMARY KEY(id))")
+
+
+async def init_complex_table() -> Session:
+    session = await init_common("CREATE TYPE IF NOT EXISTS benchmarks.udt1 (field1 text, field2 int)")
+    await session.execute(
+        "CREATE TABLE benchmarks.basic (id uuid, val int, tuuid timeuuid, ip inet, date date, time time, tuple frozen<tuple<text, int>>, udt frozen<udt1>, set1 set<int>, duration duration, PRIMARY KEY(id))"
+    )
+    return session
+
```
Lorak-mmk: I see no point in `init_common` taking schema str. You should run the execute in the caller (e.g. `init_simple_table`. You need to do this anyway for `init_complex_table`, so I see no benefit of this parameter.

benchmark/concurrent_insert.py
```diff
@@ -0,0 +1,52 @@
...
+
+from common import init_simple_table, SIMPLE_INSERT_QUERY, check_row_cnt, simple_table_cleanup, get_simple_data
+
+CONCURRENCY = 100
+
+
+async def set_up() -> tuple[Session, PreparedStatement, int]:
+    cnt = int(sys.argv[1])
+
+    session = await init_simple_table()
```
Lorak-mmk: Shouldn't concurrency be configurable?

benchmark/concurrent_ser.py
Lorak-mmk, line 1: I don't see how the serialization scenarios test serialization. They seem pretty similar to insert scenarios.
Ser / deser benchmarks should imo not execute requests, only perform ser/deser.

Lorak-mmk (changes requested):

benchmark/common.py
patrycja-ziemkiewicz: All benchmark scenarios in this PR mirror the Rust driver and Node.js driver benchmark scenarios 1:1, here are the scenarios: https://github.com/scylladb/nodejs-rs-driver/blob/main/benchmark/logic_rust/parametrized_select.rs

I did not change the benchmark structure myself, because I though We want to keep scenarios identical across all drivers.

That said, if we do not need to have exacly the same scenarios as those two drivers and want to focus mainly on comparison against the old Python driver, I can definitely make this changes.

Lorak-mmk: I get that we want to benchmark the same things, but I commented on a piece of code that I thought was not part of a benchmark itself, but benchmark preparation.
Things like keyspace / table creation, or example data insertion for SELECT benchmarks, should obviously not be measured, so there should be no problem in having differences between drivers. We do not measure them, right?

patrycja-ziemkiewicz: Unfortunately, We do measure them.  In both Node-js driver and Rust Driver this set ups are measured in the benchmarks.

patrycja-ziemkiewicz: I think I could extract set up to separate file and run it before the benchmarks to not measure it but then We would not measure same things as Node-js and Rust driver. What should I do in this situation? Also the benchmarker for now only supports running some additional commands before running benchmarks so I think We would still need to have clean up measured untill there are some changes made to the benchmarker.

Lorak-mmk: > Unfortunately, We do measure them. In both Node-js driver and Rust Driver this set ups are measured in the benchmarks.

Oh wow, this is terrible.
Let's not repeat this mistake here, do it the right way. It needs to be fixed in other benchmarks.

benchmark/concurrent_ser.py
patrycja-ziemkiewicz: I did them the same way they are done in other drivers. Should I change that?

Lorak-mmk: We do have a separate benchmarks for insert / select. When benchmarking ser / deser I see no point in also benchmarking the whole request execution. Those benchmarks should directly measure ser/deser performance, and not even require a running DB.
When improving ser/deser performance, I should be able to run the benchmarks and say "this benchmark proves that ser/deser is now X% faster". If you measure keyspace creation, request execution etc I am not able to say this. I don't even have any idea what percentage of the time is spent in ser/deser, and what in other parts of request execution.

benchmark/python-rs-driver/concurrent_insert.py
```diff
@@ -0,0 +1,46 @@
...
+    index = start_index
+
+    while index < cnt:
+        await session.execute(prepared, get_simple_data(), paged=False)
+        index += concurrency
+
+
+async def main() -> None:
+    concurrency = int(sys.argv[1])
+    cnt = int(sys.argv[2])
```
jakub-graczyk, line 29: @Lorak-mmk @patrycja-ziemkiewicz what do you think about making this configuration env variables for example SCYLLA_DRIVER_BENCHMARK_CONCURRENCY, SCYLLA_DRIVER_BENCHMARK_OP_COUNT?

This avoids confusion with config order and is more reusable imo.

benchmark/python-rs-driver/complex_select.py
```diff
@@ -0,0 +1,37 @@
...
+)
+
+
+async def test(session: Session, cnt: int, num_of_rows: int):
+    prepared = await session.prepare(COMPLEX_SELECT_QUERY)
+    for _ in range(cnt):
+        res = await session.execute(prepared, None, paged=False)
+        rows = await res.all()
+
+        print(len(rows))
```
jakub-graczyk: Is that necessary?

patrycja-ziemkiewicz: No I forget to remove that

benchmark/python-rs-driver/concurrent_insert.py
patrycja-ziemkiewicz: I believe it is better to have it as sys variable as you can set it in benchmark config and make proper use of the benchmarker.

jakub-graczyk (approved):

benchmark/python-rs-driver/select.py
```diff
@@ -0,0 +1,29 @@
...
+from scylla.session import Session
+
+from common import (
+    connect,
+    SIMPLE_SELECT_QUERY,
+)
+
+
+async def test(session: Session, cnt: int, num_of_rows: int):
+    prepared = await session.prepare(SIMPLE_SELECT_QUERY)
```
jakub-graczyk, line 12: For consistency prepare outside test(), same in complex

PR #57, opened 2026-03-17
Fixes for Statement and Prepared

paulinaczajkowska:
## Fixes for Statement and Prepared

- Added `is_serial_consistency_set` to `Statement` and `Prepared`.
- Updated `get_serial_consistency` to also return `Unset`.
- Unified getters to use `@property`, making API consistent with `Batch`.
- Added tests covering the new behavior.

## Pre-review checklist

 - [ ]  ~~I have provided docstrings for the public items that I want to introduce.~~
 - [ ]  ~~I have adjusted the documentation in `./docs/source/`.~~
 - [ ]  ~~I added appropriate `Fixes:` annotations to PR description.~~


patrycja-ziemkiewicz (commented):
Please add the row_factory parameter to the batch method docstring in session.pyi, as this PR addresses recent fixes.

src/statement.rs
```diff
@@ -197,14 +191,14 @@ impl PyStatement {
...
-        }
+        let timeout = match timeout {
+            None => Duration::MAX,
+            Some(secs) => Duration::try_from_secs_f64(secs)
+                .map_err(|_| DriverStatementConfigError::invalid_request_timeout(secs))?,
+        };
 
         let mut s = self._inner.clone();
-        s.set_request_timeout(Some(timeout.map_or(Duration::MAX, Duration::from_secs_f64)));
+        s.set_request_timeout(Some(timeout));
```
Lorak-mmk, line 252: @jakub-graczyk Please review commit "Statements: handle invalid timeout without panicking"

I think in the past we discussed some weird edge cases with timeouts and consistencies, related to issues in Rust Driver API.
I don't remember what it was about precisely, and I'm worried this commit might break some intentionally introduced workaround.

src/statement.rs
Lorak-mmk, line 1: @jakub-graczyk Ditto for commit "Statements: implement `get_serial_consistency` returning `Unset`"

src/execution_profile.rs
```diff
@@ -26,15 +26,9 @@ impl ExecutionProfile {
...
 
-        if let Some(secs) = timeout
-            && (!secs.is_finite() || secs <= 0.0)
-        {
-            return Err(DriverStatementConfigError::InvalidRequestTimeout { value: secs });
-        }
-
         if let Some(secs) = timeout {
             let duration = Duration::try_from_secs_f64(secs)
-                .map_err(|_| DriverStatementConfigError::request_timeout_conversion_failed(secs))?;
```
pasinskik, line 37: I think it's fair to get rid of `request_timeout_conversion_failed` in `execution_profile.rs` and `statement.rs`. But since that function is no longer used anywhere, it should be deleted completely from `errors.rs` as well as the enum variant `RequestTimeoutConversionFailed`.

src/batch.rs
```diff
@@ -170,16 +170,10 @@ impl PyBatch {
...
-        if let Some(secs) = timeout
-            && (!secs.is_finite() || secs <= 0.0)
-        {
-            return Err(DriverBatchError::invalid_request_timeout(secs));
-        }
-
         let timeout = match timeout {
             None => Duration::MAX,
             Some(secs) => Duration::try_from_secs_f64(secs)
-                .map_err(|_| DriverBatchError::request_timeout_conversion_failed(secs))?,
```
pasinskik, line 181: Same goes for `RequestTimeoutConversionFailed` variant of `DriverBatchError`, it's no longer needed, so it should be deleted.

src/statement.rs
jakub-graczyk: This is okay only if we allow Duration(0.0) in timeout. This was explicitly invalid before.

src/statement.rs
jakub-graczyk: This is correct. Unset means that it isn't set, None means explicitly set to None.
@paulinaczajkowska please add this to Unset docstring documenting this behaviour.

Lorak-mmk: Ok, so we can merge after the docstring is fixed.

Lorak-mmk (approved):

jakub-graczyk (approved):

PR #58, opened 2026-03-18
Expose metadata

jakub-graczyk:
This PR adds access to cluster topology and keyspace metadata from Python. Users can now inspect keyspaces, tables, columns, replication strategies, nodes, tokens, and the replica locator


src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+            _ => None,
+        }
+    }
+
+    #[getter]
+    fn other_data<'py>(&self, py: Python<'py>) -> PyResult<Option<Bound<'py, PyDict>>> {
+        match &self._inner {
+            Strategy::Other { data, .. } => Some(
+                data.iter()
+                    .map(|(key, val)| (PyString::new(py, key), PyString::new(py, val)))
```
patrycja-ziemkiewicz: Same here

wprzytula (changes requested):

paulinaczajkowska (changes requested):

src/cluster/state.rs
```diff
@@ -19,6 +23,19 @@ pub(crate) struct PyClusterState {
 
 #[pymethods]
 impl PyClusterState {
+    fn get_keyspace(&self, keyspace: String) -> Option<PyKeyspace> {
+        self._inner
+            .get_keyspace(keyspace)
+            .map(|ks| PyKeyspace { _inner: ks.clone() })
+    }
```
wprzytula: 💭 We don't need an owned `String`, do we?
If not, maybe we can borrow from `PyString` for a while, to save an allocation just for the lookup?

wprzytula (changes requested):

python/scylla/_rust/routing.pyi
```diff
@@ -12,3 +15,31 @@ class Token:
...
+        keyspace: str,
+        table: str,
+    ) -> ReplicaSet: ...
+    @property
+    def unique_nodes_in_global_ring(self) -> list[Node]: ...
+    def unique_nodes_in_datacenter_ring(self, datacenter: str) -> list[Node] | None: ...
+    @property
+    def datacenter_names(self) -> list[str]: ...
+    @property
+    def ring(self) -> list[tuple[Token, Node]]: ...
```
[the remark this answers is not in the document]
jakub-graczyk: I would change the name to `compute_*` and not make it a property in either case.

python/scylla/_rust/cluster/__init__.pyi
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Access the host ID of the node.
+        """
+        ...
+    @property
+    def shard(self) -> Shard | None:
+        """
+        Access the shard assignment of the node.
+        """
+        ...
```
wprzytula: ❓ What is _shard assignment of the node_?

wprzytula:
:question: Why is `__init__` gone from `Node` in the `pyi` file? It's still present on `NodeShard`.

wprzytula (approved):
I'm still unsure about the use of properties instead of getter methods.
Apart from this, LGTM.

python/scylla/_rust/cluster/metadata.pyi
```diff
@@ -0,0 +1,142 @@
...
+    Static = ...
+    Clustering = ...
+    PartitionKey = ...
+
+class Column:
+    @property
+    def typ(self) -> str:
+        """
+        Access the type of this column as a string.
+        """
```
Lorak-mmk: Python does have bultin `type` function, but is there any context where it would clash with a method named `type`? Code like `some_column.type()` should work.

Lorak-mmk (changes requested):

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+impl From<&ColumnKind> for PyColumnKind {
+    fn from(kind: &ColumnKind) -> Self {
+        #[deny(clippy::wildcard_enum_match_arm)]
+        match kind {
+            ColumnKind::Regular => PyColumnKind::Regular,
+            ColumnKind::Static => PyColumnKind::Static,
+            ColumnKind::Clustering => PyColumnKind::Clustering,
+            ColumnKind::PartitionKey => PyColumnKind::PartitionKey,
```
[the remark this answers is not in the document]
wprzytula: You can write something like:
`_ => unreachable!("clippy testifies that the match is exhaustive"),`

src/session.rs
```diff
@@ -102,6 +121,23 @@ impl Session {
...
+            .lock_py_attached(py)
+            .map_err(|e| PyRuntimeError::new_err(e.to_string()))?;
+        if Arc::as_ptr(&self._inner.get_cluster_state()) == Arc::as_ptr(&session_cs.get()._inner) {
+            Ok(session_cs.clone_ref(py))
+        } else {
+            *session_cs = Py::new(
+                py,
+                PyClusterState::new(self._inner.get_cluster_state(), py)?,
+            )?;
+            Ok(session_cs.clone_ref(py))
```
patrycja-ziemkiewicz: You could write it like this: 
```rust
if Arc::as_ptr(&self._inner.get_cluster_state()) != Arc::as_ptr(&session_cs.get()._inner) {
    *session_cs = Py::new(
        py,
        PyClusterState::new(self._inner.get_cluster_state(), py)?,
    )?;
}

Ok(session_cs.clone_ref(py))
```

src/cluster/state.rs
```diff
@@ -0,0 +1,130 @@
...
+        })
+    }
+}
+
+#[pymethods]
+impl PyClusterState {
+    #[getter]
+    fn get_nodes_info<'py>(&self, py: Python<'py>) -> Py<PyDict> {
+        self.known_nodes.clone_ref(py)
+    }
```
patrycja-ziemkiewicz: We are returning Python dict to a user via cloning reference. This mean that python user receives the same object, if he decide to add there some additional key (or remove the existing one) It will become present in our cache. I think it would be better to either return to user each time new python dict or some structure he cannot mutate.

src/cluster/state.rs
```diff
@@ -0,0 +1,130 @@
...
+
+#[pyclass(name = "ClusterState", frozen, skip_from_py_object)]
+pub(crate) struct PyClusterState {
+    pub(crate) _inner: Arc<ClusterState>,
+    /// Stores cached nodes as PyDict with erased type information.
+    ///
+    /// Known nodes are cached on PyClusterState creation.
+    ///
+    /// Invariant: `PyDict<Uuid, Py<PyNode>>`
+    known_nodes: Py<PyDict>,
```
patrycja-ziemkiewicz: Currently, we populate known_nodes immediately in new(). We can make this lazy by using PyOnceCell and initializing it only when the user accesses it for the first time.

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,504 @@
...
+    #[getter]
+    fn kind(&self) -> PyStrategyKind {
+        match &self._inner {
+            Strategy::SimpleStrategy { .. } => PyStrategyKind::Simple,
+            Strategy::NetworkTopologyStrategy { .. } => PyStrategyKind::NetworkTopology,
+            Strategy::LocalStrategy => PyStrategyKind::Local,
+            Strategy::Other { .. } => PyStrategyKind::Other,
+            _ => unreachable!(),
+        }
+    }
```
patrycja-ziemkiewicz: You could extract this match into `impl From<&Strategy> for PyStrategyKind`.

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,504 @@
...
+
+#[pyclass(name = "Column", frozen, skip_from_py_object)]
+#[derive(Clone)]
+pub(crate) struct PyColumn {
+    _inner: Column,
+    #[pyo3(get)]
+    typ: Py<PyString>,
+    #[pyo3(get)]
+    kind: Py<PyColumnKind>,
+}
```
patrycja-ziemkiewicz, line 128: Is there a specific reason we are keeping the _inner field here? Since we're already converting the data to Python types , _inner is currently doubling the memory usage. Do we plan to pass this back into some rust functions later?

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,504 @@
...
+    // }
+
+    // #[getter]
+    // fn partitioner<'py>(&self, py: Python<'py>) -> Option<Bound<'py, PyString>> {
+    //     self._inner
+    //         .partitioner
+    //         .as_ref()
+    //         .map(|p| PyString::new(py, p))
+    // }
+
```
patrycja-ziemkiewicz: Is this commented out on purpuse?

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,504 @@
...
+    // #[getter]
+    // fn partitioner<'py>(&self, py: Python<'py>) -> Option<Bound<'py, PyString>> {
+    //     self._inner
+    //         .view_metadata
+    //         .partitioner
+    //         .as_ref()
+    //         .map(|p| PyString::new(py, p))
+    // }
```
patrycja-ziemkiewicz: Same here

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,504 @@
...
+                Ok(s)
+            }
+        }
+    }
+
+    #[getter]
+    fn tables<'py>(&self, py: Python<'py>) -> PyResult<Py<PyDict>> {
+        if let Some(tables) = self.tables.get() {
+            return Ok(tables.clone_ref(py));
+        }
```
patrycja-ziemkiewicz: Same concern as before. We probably do not want users to mutate our cached data.

src/cluster/state.rs
```diff
@@ -36,12 +47,59 @@ impl PyClusterState {
...
+                let Some(ks) = self._inner.get_keyspace(keyspace.bind(py).to_str()?) else {
+                    return Ok(None);
+                };
+                let py_keyspace = PyKeyspace::new(py, ks.clone())?.into_py_any(py)?;
+                self.keyspaces
+                    .0
+                    .bind(py)
+                    .set_item(keyspace, py_keyspace.clone_ref(py))?;
+                Ok(Some(py_keyspace))
+            }
```
patrycja-ziemkiewicz: Same concert here. I tested it locally and user can succesfully mutate our cache or even delete some things stored there which even can create a panic for some functions for example `get_token_endpoints`. 

Image

patrycja-ziemkiewicz: This happens in many place in code right now. We should definitely change this logic in all of these places.

src/cluster/state.rs
wprzytula: Are these immutable dictionaries in Python? Such thing would fit here, right?

wprzytula: https://docs.python.org/3.15/library/stdtypes.html#frozendict

src/cluster/metadata/mod.rs
jakub-graczyk: For __repr__ it's faster to not extract stored PyStrings

src/cluster/state.rs
jakub-graczyk: https://docs.python.org/3/library/types.html#types.MappingProxyType

I'm using this one. I think it's better suited for metadata view.

python/scylla/_rust/cluster/__init__.pyi
```diff
@@ -0,0 +1,114 @@
...
+class ClusterState:
+    """
+    Represents state of the cluster allowing access to known nodes and token calculation.
+    """
+
+    @property
+    def nodes_info(self) -> MappingProxyType[UUID, Node]:
+        """
+        Access the nodes of the cluster as a dict of host ID to Node objects.
+        """
```
Lorak-mmk, line 88: Nit: it may be more idiomatic and future-proof to use `Mapping` here (I assume `MappingProxyType` implements it?).

src/cluster/node.rs
```diff
@@ -0,0 +1,101 @@
...
+            .map(|dc| PyString::new(py, dc))
+    }
+
+    #[getter]
+    fn rack<'py>(&self, py: Python<'py>) -> Option<Bound<'py, PyString>> {
+        self._inner
+            .rack
+            .as_ref()
+            .map(|rack| PyString::new(py, rack))
+    }
```
Lorak-mmk, line 64: I thought that we previously discussed also having a cache (`OnceLock` or something like that) for those.
Do I remember wrong? Or perhaps there is a reason to not do it?

src/cluster/state.rs
```diff
@@ -0,0 +1,180 @@
...
+            .map(|node| {
+                let py_node = self.get_or_init_node_from_cache(py, &node.0.host_id)?;
+                Ok::<_, PyErr>((py_node, node.1))
+            });
+        let list = PyList::empty(py);
+        for py_node in py_nodes_sequence {
+            list.append(py_node?)?;
+        }
+        Ok(list)
+    }
```
Lorak-mmk: I think there are race conditions possible, especially with GIL disabled.
Consider concurrent calls to `get_or_init_node_from_cache`.
Both calls can get `None` from `get_item`, create new node object, and put it int the dict. Now you got 2 Python Node objects for the same Rust Node objects.

Your `cache_flag` check in `get_or_init_node_from_cache` also looks problematic. A call to `get_or_init_node_from_cache` could get `None` from `get_item`, then a call to `get_nodes_info` could fill all the items, then your `cache_flag.load` return True, and falsely return `Ok(None)` to the user.

src/cluster/state.rs
```diff
@@ -0,0 +1,180 @@
...
+    ) -> PyResult<Option<Py<PyAny>>> {
+        let cache = self.known_nodes.0.bind(py);
+        let cache_flag = &self.known_nodes.1;
+        match cache.get_item(node_id)? {
+            Some(node) => Ok(Some(node.unbind().clone_ref(py))), // Node present in cache
+            None => {
+                if cache_flag.load(std::sync::atomic::Ordering::Relaxed) {
+                    // O(1) return when node is not in cache and all nodes are known.
+                    return Ok(None);
+                }
```
Lorak-mmk: This is a private function that is always called with "correct" uuid, right? By correct I mean that it is a Uuid always present in ClusterState.
If so, it should not be possible for the node to be missing, and so this case implies a bug in the driver. Returning `Ok(None)` is not right.

Lorak-mmk (changes requested):
Posting review of first 4 commits, because I believe the caching implementation here has serious issues .

python/scylla/_rust/session.pyi
```diff
@@ -10,6 +11,12 @@ class Session:
     Represents a CQL session, which can be used to communicate with the database.
     """
 
+    @property
+    def cluster_state(self) -> ClusterState:
+        """
+        Access information about the cluster topology or schema through ClusterState object.
+        """
+        ...
```
Lorak-mmk, line 19: In this specific case, I think explicit getter would be better than a property. Cluster state may change (be replaced), and it is imo less obvious when it is exposed as a property.
Let's use properties for stuff that is immutable, and cheap to compute (caching is fine).

python/scylla/_rust/cluster/metadata.pyi
```diff
@@ -0,0 +1,137 @@
...
+        ...
+    @property
+    def kind(self) -> ColumnKind:
+        """
+        Access the kind of this column.
+        """
+        ...
+    def __repr__(self) -> str: ...
+
```
Lorak-mmk, line 134: I think we discussed renaming `typ` to `type` here.

src/cluster/metadata/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        for (name, column) in &self._inner.columns {
+            match cache.get_item(name)? {
+                Some(_) => {} // Column is present in cache
+                None => {
+                    let py_column = PyColumn::new(column.clone(), py)?.into_py_any(py)?;
+                    cache.set_item(name, py_column)?;
+                }
+            }
+        }
```
Lorak-mmk: Please try to think of some reusable abstraction for cache. There should be no need to understand the same atomic-and-dict logic in each type.

python/scylla/_rust/cluster/metadata.pyi
Lorak-mmk, line 134: Huh, I wanted to say that a Column should also have a name, not only type and kind, but I see this is an artifact of Rust Driver API, in which Column also doesn't store its name.

python/scylla/_rust/cluster/metadata.pyi
```diff
@@ -0,0 +1,137 @@
...
+    Clustering = ...
+    PartitionKey = ...
+
+class Column:
+    @property
+    def typ(self) -> str:
+        """
+        Access the type of this column as a string.
+        """
+        ...
```
Lorak-mmk, line 126: Why `typ` returns a string? Can't we have some type analogous to `ColumnType` from Rust Driver?

Lorak-mmk (changes requested):

python/scylla/_rust/cluster/metadata.pyi
jakub-graczyk: Reason for `typ` is exactly the same as for Rust Driver, `type` is a keyword.

src/cluster/metadata/column_type.rs
patrycja-ziemkiewicz, line 1: All of the classes defined here should be `frozen`

src/cluster/metadata/column_type.rs
```diff
@@ -0,0 +1,474 @@
...
+}
+
+#[pyclass(name = "CqlBoolean", extends=PyCqlNativeType)]
+pub(crate) struct PyCqlBoolean;
+
+impl PyCqlBoolean {
+    pub(crate) fn new() -> PyClassInitializer<Self> {
+        PyCqlNativeType::new().add_subclass(Self {})
+    }
+}
```
patrycja-ziemkiewicz: I believe (unfortunately) that it would be more beneficial to write macro for this native classes.

src/cluster/metadata/column_type.rs
```diff
@@ -0,0 +1,474 @@
...
+
+#[pyclass(name = "CqlVector", extends=PyCqlColumnType)]
+pub(crate) struct PyCqlVector {
+    #[pyo3(get)]
+    pub(crate) typ: Py<PyCqlColumnType>,
+    #[pyo3(get)]
+    pub(crate) dimensions: u16,
+}
```
patrycja-ziemkiewicz: instead of `#[pyo3(get)]` for every field in struct you could use `#[pyclass(get_all)]`

src/cluster/metadata/column_type.rs
```diff
@@ -0,0 +1,474 @@
...
+pub(crate) struct PyCqlUserDefinedType {
+    #[pyo3(get)]
+    pub(crate) name: String,
+    #[pyo3(get)]
+    pub(crate) frozen: bool,
+    #[pyo3(get)]
+    pub(crate) keyspace: String,
+    #[pyo3(get)]
+    pub(crate) field_types: Vec<(String, Py<PyCqlColumnType>)>,
+}
```
patrycja-ziemkiewicz: Use `#[pyclass(get_all)]` instead of ` #[pyo3(get)]`.

src/cluster/metadata/column_type.rs
```diff
@@ -0,0 +1,474 @@
...
+                        .expect("this extract can't fail")
+                }),
+                NativeType::Boolean => Py::new(py, PyCqlBoolean::new()).map(|obj| {
+                    obj.extract::<Py<PyCqlColumnType>>(py)
+                        .expect("this extract can't fail")
+                }),
+                NativeType::Counter => Py::new(py, PyCqlCounter::new()).map(|obj| {
+                    obj.extract::<Py<PyCqlColumnType>>(py)
+                        .expect("this extract can't fail")
+                }),
```
patrycja-ziemkiewicz: We should not call `expect` here, `Py::new` migh fail from various different reasons for example when encountering out of memory error.

src/cluster/metadata/column_type.rs
```diff
@@ -0,0 +1,474 @@
...
+    #[deny(clippy::wildcard_enum_match_arm)]
+    match column_type {
+        ColumnType::Native(native_type) =>
+        {
+            #[deny(clippy::wildcard_enum_match_arm)]
+            match native_type {
+                NativeType::Ascii => Py::new(py, PyCqlAscii::new()).map(|obj| {
+                    obj.extract::<Py<PyCqlColumnType>>(py)
+                        .expect("this extract can't fail")
+                }),
```
patrycja-ziemkiewicz: I think it would be more efficient to use `cast`  here.

jakub-graczyk: You can't cast to T without Py< T> . And you can't call `.cast()` on the PyClassInitializer without the Py<>, `extract()` is necessary here.

src/cluster/metadata/column_type.rs
jakub-graczyk: This is not what the code does. `expect()` is called on the extract which can't fail because it's inside `map` so `Py::new()` returned Ok() and extract can't fail here because we are extracting parent from subclass (which was successfully created). `Py::new()` error is propagated correctly.

src/cluster/metadata/column_type.rs
patrycja-ziemkiewicz: No it is not. You could convert value `into_bound` and then call `cast_into`. However I found even better method for this use case:
image

src/cluster/node.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    #[getter]
+    fn address(&self) -> (IpAddr, u16) {
+        (self._inner.address.ip(), self._inner.address.port())
+    }
+
+    #[getter]
+    fn datacenter<'py>(&self, py: Python<'py>) -> Option<Py<PyString>> {
+        self.datacenter
+            .get_or_init_py_attached(py, || {
```
wprzytula, line 43: ❓ Have we forbidden `read`, `write`, `get_or_init` (without `_py_attached`) in clippy config? I recall we decided it's the proper way to prevent bugs due to not ensuring GIL is held.

src/cluster/node.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl From<Arc<Node>> for PyNode {
+    fn from(_inner: Arc<Node>) -> Self {
+        Self {
+            _inner,
+            datacenter: OnceLock::new(),
+            rack: OnceLock::new(),
+        }
+    }
+}
```
wprzytula, line 26: 💭  Did we discuss and decide that lazy approach here is worth it? It complicates code quite a lot.

src/cluster/state.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn __repr__<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyString>> {
+        PyString::from_fmt(
+            py,
+            format_args!(
+                "ClusterState(nodes={}, keyspaces={})",
+                self._inner.get_nodes_info().len(),
+                self._inner.keyspaces_iter().count()
+            ),
+        )
+    }
```
wprzytula, line 161: 🤔 Is it suitable for `__repr__` impl to print just number of elements in the contained lists?

src/errors.rs
```diff
@@ -417,11 +417,17 @@ impl From<DriverDeserializationError> for PyErr {
 #[must_use]
 pub enum DriverSessionConnectionError {
     /// The Tokio task running session creation failed to join.
-    RuntimeTaskJoinFailed { message: String },
+    RuntimeTaskJoinFailed {
+        message: String,
+    },
     /// The Rust driver failed to establish a new session.
     NewSessionError {
         source: Box<scylla::errors::NewSessionError>,
```
wprzytula, line 425: ❓ Isn't this change enforced by `cargo fmt`? Doesn't this mean that this should belong to an earlier commit?

src/errors.rs
```diff
@@ -436,6 +442,10 @@ impl DriverSessionConnectionError {
             source: Box::new(source),
         }
     }
+
+    pub(crate) fn python_conversion_error(source: PyErr) -> Self {
+        Self::PythonConversionError { source }
+    }
```
wprzytula, line 448: ❓ What's the point of this function over using explicit constructor `DriverSessionConnectionError::PythonConversionError`?

src/cluster/state.rs
```diff
@@ -18,6 +18,29 @@ pub(crate) struct PyClusterState {
...
+            })?;
+
+            Ok(())
+        })?;
+        Ok(Self {
+            _inner: inner,
+            known_nodes: nodes_cache,
+        })
+    }
+}
```
wprzytula, line 43: ❓ Please confirm that I undestand this correctly.

1. Rust collection of PyNodes (HashMap) is created eagerly.
2. Each PyNode is created eagerly while creating the HashMap.
3. PyNode itself has lazily initialized properties.
4. Python-side map (mirroring the HashMap) is initialized lazily.

src/session.rs
```diff
@@ -24,6 +26,20 @@ use std::future::Future;
 #[derive(Clone)]
 pub(crate) struct Session {
     pub(crate) _inner: Arc<ScyllaSession>,
+    pub(crate) cluster_state: Arc<Mutex<Py<PyClusterState>>>,
+}
```
wprzytula, line 30: 🤔 Is it correct and optimal to have two distinct Arcs for `ScyllaSession` and `cluster_state`? Shouldn't they be put in the same `Arc`? Can they be managed separately, or always together?

wprzytula, line 30: 💭 That's a lot of indirection:
1. `Session` is passed by (GC-handled) reference in Python.
2. It contains `Arc` in `cluster_state`.
3. Then there's a `Py` reference-counted smart pointer inside.
4. And again `Arc` inside.

❓ Are all these layers really necessary? Please explain them and justify need for each.

src/session.rs
```diff
@@ -110,6 +126,19 @@ impl Session {
...
+        let mut session_cs = self.cluster_state.lock_py_attached(py).unwrap();
+        if Arc::as_ptr(&self._inner.get_cluster_state()) != Arc::as_ptr(&session_cs.get()._inner) {
+            *session_cs = Py::new(
+                py,
+                PyClusterState::try_from(self._inner.get_cluster_state())?,
+            )?;
+        }
+
+        Ok(session_cs.clone_ref(py))
+    }
```
wprzytula, line 150: 🔧 Write comments explaining this logic in detail.

wprzytula: In general, comments (even redundant!) would help me navigate in this code. The same goes for future readers, I suppose.

src/session_builder.rs
```diff
@@ -103,9 +103,8 @@ impl SessionBuilder {
             .spawn(async move { scylla::client::session::Session::connect(config).await })
             .await?;
         match session_result {
-            Ok(session) => Ok(Session {
-                _inner: Arc::new(session),
-            }),
+            Ok(session) => Session::try_from(Arc::new(session))
+                .map_err(DriverSessionConnectionError::python_conversion_error),
```
wprzytula: ⛏️ Here the associated function really gives nothing over the plain enum variant constructor.

python/scylla/_rust/cluster/metadata.pyi
wprzytula, line 1: ♻️ `add: metadata module` commit is very big. Please split it into logical chunks.

src/routing/mod.rs
```diff
@@ -40,8 +48,202 @@ impl PyToken {
...
+            return Ok(None);
+        };
+        let py_cs = self._inner.bind(py).get();
+        let result = py_cs.known_nodes.get_or_init(py, &node.host_id, |key| {
+            py_cs
+                ._inner
+                .known_peers
+                .get(key)
+                .map(|node| Py::new(py, PyNode::from(node.clone())))
+                .transpose()
```
wprzytula: ⛏️ Use `Arc::clone(node)` instead of `node.clone()`.

src/routing/mod.rs
```diff
@@ -40,8 +48,202 @@ impl PyToken {
...
+                .known_nodes
+                .get_or_init(py, &node.host_id, |key| {
+                    py_cs
+                        ._inner
+                        .known_peers
+                        .get(key)
+                        .map(|node| Py::new(py, PyNode::from(node.clone())))
+                        .transpose()
+                })?
+                .expect("node can't be known by Rust Driver and simultaneously None");
```
wprzytula: ♻️ This part is duplicated in at least two places. Let's extract it somehow.

wprzytula (changes requested):

src/cluster/node.rs
jakub-graczyk: It requires refactoring code not relevant to metadata, I decided to leave it out of this (already big) PR.

src/cluster/node.rs
jakub-graczyk: Earlier comment by @Lorak-mmk :
I thought that we previously discussed also having a cache (`OnceLock` or something like that) for those.
Do I remember wrong? Or perhaps there is a reason to not do it?

src/errors.rs
jakub-graczyk: This reformatting is applied only when I add a new variant so I believe it belongs here.

src/errors.rs
jakub-graczyk: Consistency with existing errors for failing python conversion.

src/cluster/state.rs
jakub-graczyk: 1. yes
2. yes
3. yes
4. it is also created eagerly here, Py<> returned to the Python Mapping is discarded here.

There are 2 approaches:
1. Either eager (above) initialization with cache freezing ensuring all nodes for this snapshot of cluster state are created with a new `PyClusterState` instance.
2. They are created lazily when called by the method.

Main difference aside from lazy vs. eager discussion is Replica Locator.
When accessing replicas we need to provide an init function to `get_or_init()` a node from cache. If approach 1. is taken the closure can be `|| unreachable!()` because the cache is frozen and all nodes are known the closure will never be called.

Or approach 2. which requires for each node access in replica locator to specify a closure that creates a node in  `get_or_init()`.

Currently I use approach 1. and provide a correct closure which is never called due to freezing the node cache when creating `PyClusterState`.

I prefer approach 1 with  `|| unreachable!()` because there aren't usually "a lot" of nodes + when accessing cluster state nodes will probably be accessed anyway.

src/session.rs
jakub-graczyk: When cluster state is stale when returned from `get_cluster_state()` it's swapped out with a new `Py<PyClusterState>` while `Arc<ScyllaSession>` stays the same.

jakub-graczyk: Each pyclass must be Send + Sync. for interor mutability we need `Arc<Mutex<<T>>`. 

Py<T> is here because we wan't to return the same object (`clone_ref()`) when returning the cluster state and prevent it from being freed. As earlier mentioned at any moment `cluster_state` field could be "stale" only updated with a new accurate copy on  `get_cluster_state()`.

`PyClusterState` itself holds an Arc to cluster state to force Rust driver to allocate a different memory location for a new cluster state in order to differentiate by comparing pointers when the cluster state changes.

python/scylla/_rust/cluster/metadata.pyi
jakub-graczyk: There is barely any logic in cloumn_type.rs, and mod.rs has already been reviewed.

src/routing/mod.rs
jakub-graczyk: This can be replaced with `|| unreachable!()`  if the eager approach is taken.

src/cluster/node.rs
wprzytula: Do we have an issue for that? If not, please open one.

src/cluster/node.rs
wprzytula: > It complicates code quite a lot.

I'm just asking if we had a conscious design decision that we would use lazy approach here because it outweighs the costs.

src/session.rs
wprzytula: Is it really? I would think that the `Mutex's` contents are swapped out, not the `Arc<Mutex>` itself.

wprzytula: I suggest holding `Arc<(ScyllaSession, Mutex<Py<PyClusterState>>)>`. This could be more elegant if the tuple is replaced with e.g. `SessionInner` named field struct.

wprzytula: > Each pyclass must be Send + Sync.

Which indirection layer does this justify?

> for interor mutability we need `Arc<Mutex<<T>>`.

AFAIU, `Mutex<T>` is needed for interior mutability, not `Arc<Mutex<T>>`.
 
> Py is here because we wan't to return the same object (`clone_ref()`) when returning the cluster state and prevent it from being freed. As earlier mentioned at any moment `cluster_state` field could be "stale" only updated with a new accurate copy on `get_cluster_state()`.

I think I get it.

> `PyClusterState` itself holds an Arc to cluster state to force Rust driver to allocate a different memory location for a new cluster state in order to differentiate by comparing pointers when the cluster state changes.

This part is OK to me, because Rust Driver's API anyway returns `Arc<ClusterState>`, shared with all other users of the `ArcSwap` shared by `ClusterWorker` and `Cluster`.

src/routing/mod.rs
wprzytula: If it's not replaced with `unreachable!()`, please deduplicate.

src/cluster/state.rs
wprzytula: > 4\. Py<> returned to the Python Mapping is discarded here.

WDYM? Where?

python/scylla/_rust/cluster/metadata.pyi
wprzytula: Why not split? I'm sure it would improve the quality and ease of the review, at least from my side.

src/cluster/node.rs
jakub-graczyk: #64

src/session.rs
jakub-graczyk: This will be addressed with #78 and #79. Right now it's not ergonomic to hold it inside 1 `Arc` , `Arc` around `Mutex` will be removed when `#[derive(Clone)]` is dropped with #79.

src/session.rs
```diff
@@ -110,9 +126,31 @@ impl Session {
...
+        // we can determine if the `PyClusterState`'s snapshot is stale
+        // and needs to be replaced with a fresh snapshot.
+        let mut py_cluster_state = self.cluster_state.lock_py_attached(py).unwrap();
+        let rust_current_cluster_state = self._inner.get_cluster_state();
+        let python_snapshot_cluster_state = &py_cluster_state.get()._inner;
+        if Arc::as_ptr(&rust_current_cluster_state) != Arc::as_ptr(python_snapshot_cluster_state) {
+            *py_cluster_state = Py::new(
+                py,
+                PyClusterState::try_from(self._inner.get_cluster_state())?,
+            )?;
```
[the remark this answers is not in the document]
jakub-graczyk: This is intentional

python/scylla/_rust/cluster/__init__.pyi
```diff
@@ -0,0 +1,100 @@
...
+        ...
+    @property
+    def enabled(self) -> bool:
+        """
+        Is `True` if the node is enabled, `False` otherwise.
+
+        Only enabled nodes will have connections open.
+        """
+        ...
```
wprzytula, line 63: 🔧 Let's explain what it means that a node is enabled, and under what circumstances it can be disabled.

src/session.rs
```diff
@@ -24,6 +26,20 @@ use std::future::Future;
...
+    type Error = PyErr;
+    fn try_from(_inner: Arc<Session>) -> Result<Self, Self::Error> {
+        let cluster_state = Python::attach(|py| {
+            Py::new(py, PyClusterState::try_from(_inner.get_cluster_state())?)
+        })?;
+        Ok(Self {
+            cluster_state: Arc::new(Mutex::new(cluster_state)),
+            _inner,
+        })
+    }
```
wprzytula, line 42: ❓ Is there any reason why you decided to use `TryFrom` instead of a plain `PySession::new(session: Arc<Session>)` constructor? `From`/`TryFrom` impls tend to be less discoverable than constructors.

wprzytula (approved):

src/utils.rs
```diff
@@ -25,13 +29,14 @@ pub(crate) fn add_submodule(
...
+    let full_name = format!("{}.{name}", parent_mod.name()?);
+    let sub_module = PyModule::new(py, &full_name)?;
+    module_constructor(py, &sub_module)?;
     parent_mod.add_submodule(&sub_module)?;
     py.import("sys")?
         .getattr("modules")?
-        .set_item(format!("{}.{name}", parent_mod.name()?), sub_module)?;
+        .set_item(&full_name, sub_module)?;
     Ok(())
 }
```
Lorak-mmk, line 42: Commit: "fix: submodule creation"

Nice catch! Hopefully later commits introduce some tests for that. I remember that in the poc I wrote some tests for importing. 
They could be extended after you introduce such nested module.

src/session.rs
```diff
@@ -110,6 +126,28 @@ impl PySession {
...
+        // and needs to be replaced with a fresh snapshot.
+        let mut py_cluster_state = self.cluster_state.lock_py_attached(py).unwrap();
+        let rust_current_cluster_state = self._inner.get_cluster_state();
+        let python_snapshot_cluster_state = &py_cluster_state.get()._inner;
+        if Arc::as_ptr(&rust_current_cluster_state) != Arc::as_ptr(python_snapshot_cluster_state) {
+            *py_cluster_state = Py::new(
+                py,
+                PyClusterState::try_from(self._inner.get_cluster_state())?,
+            )?;
+        }
```
Lorak-mmk, line 147: There is https://doc.rust-lang.org/std/sync/struct.Arc.html#method.ptr_eq for that, could you please switch to it?

Lorak-mmk (changes requested):

src/utils.rs
Lorak-mmk: Now I see there are no such tests. Please add some to my import tests.

python/scylla/_rust/cluster/__init__.pyi
wprzytula: This should also be improved in the Rust Driver, as we discussed lastly.

src/session.rs
jakub-graczyk: Discussed in a meeting.

Lorak-mmk (approved):

Issue #6, opened 2025-09-15
Decide on SessionBuilder API


Lorak-mmk:

wprzytula:
Would each method call mutate the builder instance and return it?

PR #60, opened 2026-03-27
SessionBuilder: introduce builder-style configuration and custom strategy methods

patrycja-ziemkiewicz:
This PR is a smaller part of the broader SessionBuilder/configuration API work, split out to keep the changes easier to review.

Changes included in this PR:
- redesigned session configuration API to use the builder pattern instead of constructor-based configuration,
- added SessionBuilder support for custom authentication,
- added SessionBuilder support for custom address translator,
- added SessionBuilder support for custom timestamp generator,
- added SessionBuilder support for custom host filter.

This PR contains all builder methods that require users to provide their own custom Python class implementations, except for TLS, which was discussed to be handled in a separate PR.

Ref: https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/6

## Pre-review checklist


python/scylla/_rust/session_builder.pyi
```diff
@@ -1,10 +1,81 @@
+from __future__ import annotations
+
 from collections.abc import Sequence
+from ipaddress import IPv4Address, IPv6Address
 
 from .execution_profile import ExecutionProfile
 from .session import Session
 
+ContactPoint = str | tuple[str, int] | tuple[IPv4Address | IPv6Address, int]
```
wprzytula: ❓ Why not this way?

```suggestion
ContactPoint = str | tuple[str | IPv4Address | IPv6Address, int]
```

python/tests/test_session_builder.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_custom_host_filter_success() -> None:
+    host_filter = AcceptAllHostFilter()
+
+    builder = (
+        SessionBuilder().contact_points([("127.0.0.2", 9042)]).user("cassandra", "cassandra").host_filter(host_filter)
+    )
```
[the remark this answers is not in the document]
Lorak-mmk: Don't we have some autoformatter for Python? Does it not touch this dir, or does it allow such long lines?

src/session_builder.rs
```diff
@@ -68,6 +117,106 @@ impl SessionBuilder {
...
+    type Error = PyErr;
+
+    fn extract(obj: Borrowed<'_, 'py, PyAny>) -> Result<Self, Self::Error> {
+        if let Ok(s) = obj.extract::<ContactPoint>() {
+            return Ok(ContactPoints::Single(s));
+        }
+
+        if let Ok(l) = obj.cast::<PySequence>() {
+            return Ok(ContactPoints::Multiple(l.to_owned().unbind()));
+        }
```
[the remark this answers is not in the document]
Lorak-mmk: This comment seems important. Is it correct?

python/tests/test_session_builder.py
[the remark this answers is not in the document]
patrycja-ziemkiewicz: It allows such long lines. It is configured for line-length = 120. I can make it accept line-length = 88.

src/session_builder.rs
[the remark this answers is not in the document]
patrycja-ziemkiewicz: No it is not. In this example extraction of ContactPoint will fail as we only accept tuple(str, int) or tuple(ipaddress, int) so here the extraction will fails and We will be able to extract PySequence.

python/tests/test_session_builder.py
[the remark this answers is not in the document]
Lorak-mmk: no need to change, 120 is fine. This line just looked longer, but I see it is 118

src/session_builder.rs
```diff
@@ -68,6 +58,105 @@ impl SessionBuilder {
...
+            return Ok(ContactPoint::Host(s));
+        }
+
+        if let Ok(tuple) = obj.cast::<PyTuple>() {
+            if tuple.len() != 2 {
+                return Err(PyErr::new::<PyTypeError, _>(format!(
+                    "Invalid tuple length: expected 2, got {}",
+                    tuple.len()
+                )));
+            }
```
wprzytula: 🔧  Error message should provide the context, such as you wrote below:
> "Invalid contact point type: ..."

src/session_builder.rs
```diff
@@ -68,6 +58,105 @@ impl SessionBuilder {
...
+                for item in seq.try_iter()? {
+                    let item = item?;
+                    item.extract::<ContactPoint>()?.add_known_node(config);
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
```
wprzytula, line 168: 💭 Will errors propagated here using `?` contain enough context for the user? Please post example error messages emitted here.

python/tests/helpers/ccm.py
wprzytula, line 1: ♻️ Introducing these CCM helpers should constitute a separate commit.

wprzytula: Or even a separate PR, but I won't insist.

src/policies.rs
```diff
@@ -0,0 +1,102 @@
+use async_trait::async_trait;
+use pyo3::exceptions::PyRuntimeError;
+use pyo3::prelude::{PyAnyMethods, PyModule, PyModuleMethods};
+use pyo3::types::{PyDict, PyTuple};
+use pyo3::{Bound, Py, PyResult, Python, pyclass, pymethods, pymodule};
+use scylla::authentication::{AuthError, AuthenticatorProvider, AuthenticatorSession};
+//TODO:
+// Ask if We want to implement some of the Custom Authenticator that are implemented
+// in python driver. This would improve performance for python users that would
+// want to use them, as it would eliminate the need to call python code from rust for each authentication step.
```
wprzytula: If you ask me, this is not a priority. Let's have an issue about that, that's it for now.

wprzytula (changes requested):

python/scylla/_rust/session_builder.pyi
Lorak-mmk: +1

src/session_builder.rs
```diff
@@ -68,6 +58,105 @@ impl SessionBuilder {
...
+            if let Ok(host) = host.extract::<IpAddr>() {
+                return Ok(ContactPoint::SocketAddr(SocketAddr::new(host, port)));
+            }
+        }
+
+        Err(PyErr::new::<PyTypeError, _>(format!(
+            "Invalid contact point type: expected str | tuple(str, int) | tuple(ipaddress, int), got {}",
+            obj.get_type().name()?
+        )))
+    }
```
Lorak-mmk: Nit: let-else would fit much better here, because it would avoid this big indented block.

src/session_builder.rs
Lorak-mmk: I think that Rust tuples up to certain length implement `FromPyObject`, perhaps you could use it here?

src/session_builder.rs
```diff
@@ -68,6 +58,105 @@ impl SessionBuilder {
...
+                return if let Ok(ip) = IpAddr::from_str(host_str) {
+                    Ok(ContactPoint::SocketAddr(SocketAddr::new(ip, port)))
+                } else {
+                    Ok(ContactPoint::Host(format!("{}:{}", host_str, port)))
+                };
+            }
+
+            if let Ok(host) = host.extract::<IpAddr>() {
+                return Ok(ContactPoint::SocketAddr(SocketAddr::new(host, port)));
+            }
```
Lorak-mmk: Hmm... you probably don't use it because you wanted to be able to try 2 different types. I think a tuple like `(Bound<'py, PyAny>, u16)` should still work.
Or perhaps this whole block could use that, like
```rust
if let Ok((host_str, port)) = obj.extract::<(&str, u16)>() {
    todo!()
}
if let Ok((host, port)) = obj.extract::<(IpAddr, u16)>() {
    todo!()
}
...

File: python/tests/helpers/ccm.py
Comment (Lorak-mmk) [inline python/tests/helpers/ccm.py:1]: Adding CCM should definitely be in a separate commit.

File: src/policies.rs (hunk 6 above)
Comment (Lorak-mmk) [inline src/policies.rs]: Open an issue about that, but don't implement them now.

File: src/policies.rs [hunk 9]
```diff
@@ -0,0 +1,102 @@
...
+
+    fn evaluate_challenge(&self, _challenge: Option<&[u8]>) -> PyResult<Option<Vec<u8>>> {
+        Err(PyRuntimeError::new_err("Method unimplemented"))
+    }
+
+    fn success(&self, _token: Option<&[u8]>) -> PyResult<()> {
+        Ok(())
+    }
+}
+
```
Lorak-mmk: Shouldn't all methods here return `Err(PyRuntimeError::new_err("Method unimplemented"))`?

python/scylla/_rust/policies.pyi
```diff
@@ -0,0 +1,21 @@
+from typing import Any, Optional
+
+class Authenticator:
+    """
+    Base class for implementing custom authentication logic.
+
+    Users should subclass this and override the methods to interface with
+    custom auth providers (e.g., LDAP, Kerberos).
+    """
+    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
```
Lorak-mmk, line 13: `Authenticator` is supposed to be an abstract class. Users should subclass it and make instances of their own subclasses. Why do we need `__init__`?

Note: It is a genuine question, I'm not saying it is wrong. There may be a reason I just don't see.

src/policies.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn evaluate_challenge(&self, _challenge: Option<&[u8]>) -> PyResult<Option<Vec<u8>>> {
+        Err(PyRuntimeError::new_err("Method unimplemented"))
+    }
+
+    fn success(&self, _token: Option<&[u8]>) -> PyResult<()> {
+        Ok(())
+    }
+}
```
Lorak-mmk, line 31: In Rust Driver we have 2 traits (as you surely know because you implemented them).

First we start authentication session. This corresponds to [AUTHENTICATE](https://github.com/apache/cassandra/blob/88e3fcfe8b436b2adf8f686d2e86e5bb9a94b50b/doc/native_protocol_v4.spec#L517-L534) cql message received from the server.
`start_authentication_session` on `AuthenticatorProvider` returns auth session object, and a payload that will be sent to a server in [AUTH_RESPONSE](https://github.com/apache/cassandra/blob/88e3fcfe8b436b2adf8f686d2e86e5bb9a94b50b/doc/native_protocol_v4.spec#L311-L328)

After that the authentication session may continue, by server sending us [AUTH_CHALLENGE](https://github.com/apache/cassandra/blob/88e3fcfe8b436b2adf8f686d2e86e5bb9a94b50b/doc/native_protocol_v4.spec#L802-L812) messages, to which we again response with `AUTH_RESPONSE`.

Session finished  with server sending us either [AUTH_SUCCESS](https://github.com/apache/cassandra/blob/88e3fcfe8b436b2adf8f686d2e86e5bb9a94b50b/doc/native_protocol_v4.spec#L814-L822) or ERROR.

Now, the problem in your code is that the authentication process is multi-step, and may be stateful. Responses we send in each step may depend on data received in previous steps.
In your implementation, a single object is both the authenticator (session-factory) and a session. It has no way to be stateful and produce correct responses in multi-step scenarios. We need 2 separate objects.

pyproject.toml
```diff
@@ -37,6 +38,9 @@ cache-keys = [
 ]
 config-settings = { build-args = '--profile=dev' }
 
+[tool.uv.sources]
+ccm = { git = "https://github.com/scylladb/scylla-ccm.git", branch = "master" }
+
```
Lorak-mmk: Can't this be expressed in `dependencies` table? We do it in Python Driver https://github.com/scylladb/python-driver/blob/caa98b60998729b97ad1a375ea1c7723e1728161/pyproject.toml#L57 so why not here?

src/policies.rs
```diff
@@ -95,8 +98,90 @@ impl AuthenticatorSession for InternalAuthenticator {
...
+
+#[pymethods]
+impl PyAddressTranslator {
+    #[expect(unused_variables)]
+    #[new]
+    #[pyo3(signature = (*args, **kwargs))]
+    pub fn new(args: &Bound<'_, PyTuple>, kwargs: Option<&Bound<'_, PyDict>>) -> Self {
+        PyAddressTranslator {}
+    }
+
```
Lorak-mmk, line 143: ditto, why do we need to be able to create an abstract class? PyAddressTranslator in this case.

src/policies.rs
```diff
@@ -178,10 +180,52 @@ impl From<&UntranslatedPeer<'_>> for PyPeerInfo {
...
+pub(crate) struct PyTimestampGenerator {}
+
+#[pymethods]
+impl PyTimestampGenerator {
+    #[expect(unused_variables)]
+    #[new]
+    #[pyo3(signature = (*args, **kwargs))]
+    pub fn new(args: &Bound<'_, PyTuple>, kwargs: Option<&Bound<'_, PyDict>>) -> Self {
+        PyTimestampGenerator {}
+    }
```
Lorak-mmk, line 222: ditto abut creation, this time for timestamp generator.

python/scylla/_rust/policies.pyi
Lorak-mmk, line 1: Commit: "Add timestamp generator to SessionBuilder"

Please look at the commit message body, it looks like a mistake.

python/scylla/_rust/policies.pyi
```diff
@@ -19,3 +22,31 @@ class Authenticator:
...
+
+    @property
+    def host_id(self) -> uuid.UUID: ...
+    @property
+    def address(self) -> Tuple[ipaddress.IPv4Address | ipaddress.IPv6Address, int]: ...
+    @property
+    def datacenter(self) -> Optional[str]: ...
+    @property
+    def rack(self) -> Optional[str]: ...
+    def __repr__(self) -> str: ...
```
Lorak-mmk: In your PR you have PeerInfo for AddressTranslator, and Peer for HostFilter.
In Rust Driver we have Peer for HostFilter, and UntranslatedPeer for AddressTranslator.

It is not at all obvious what is the difference between PeerInfo and Peer, rust Driver naming is imo better. Also, for PeerInfo (=UntranslatedPeer), In Rust Driver we have `untranslated_address` field/method. We might want to do the same here.

src/policies.rs
```diff
@@ -221,11 +223,89 @@ impl TimestampGenerator for InternalTimestampGenerator {
...
+
+#[pymethods]
+impl PyHostFilter {
+    #[expect(unused_variables)]
+    #[new]
+    #[pyo3(signature = (*args, **kwargs))]
+    pub fn new(args: &Bound<'_, PyTuple>, kwargs: Option<&Bound<'_, PyDict>>) -> Self {
+        PyHostFilter {}
+    }
+
```
Lorak-mmk, line 264: and a final ditto about creation, this time for host filter.

Lorak-mmk (changes requested):

src/policies.rs
patrycja-ziemkiewicz: I was copying this logic from old python driver where thay have default behaviour for this two methods. 
image

Lorak-mmk: Perhaps Python Driver always sets an authenticator? If user didn't provide one, then this default one is used?
Please check if this is the case. If it is, it explains Python Drivers behavior - it wants default authenticator to be a reasonable placeholder .

In Rust Driver it is different: authenticator is Option on SessionConfig, so you don't need a default one.

python/scylla/_rust/policies.pyi
patrycja-ziemkiewicz: In PyO3, even 'abstract' base classes require a #[new] method so that Python can allocate the underlying Rust struct when a subclass is instantiated.

We use *args and **kwargs in the signature to enable the constructor to have arguments. This allows users to define custom __init__ methods with any arguments without triggering a signature mismatch error in the Rust-side allocation.

src/session_builder.rs
patrycja-ziemkiewicz: I created test case that checks if the correct error message is propagated. In test_session_builder.py file you can find this example and emitted error message content.

src/policies.rs
patrycja-ziemkiewicz: In python driver if no authenticator is provided it sets the value to PlainTextAuthProvider, which extends the base class Authenticator.
image

src/session_builder.rs
```diff
@@ -82,6 +53,94 @@ impl SessionBuilder {
...
+            return Ok(ContactPoints::Single(s));
+        }
+
+        if let Ok(l) = obj.cast::<PySequence>() {
+            return Ok(ContactPoints::Multiple(l.to_owned().unbind()));
+        }
+
+        Err(DriverSessionConfigError::contact_points_type_error(obj))
+    }
+}
```
Lorak-mmk, line 201: For a single contact point you convert it fully in `FromPyObject`, but for multiple contact points you only convert it to `PySequence`. The rest of the conversion is then done in `add_known_nodes`.

I guess you did that to avoid allocating a Vec of contact points. Imo this is not a good choice here. Creating a session is not a hot path, and we shouldn't worry about single allocation here. Instead we should focus on writing clear code. Let's have full conversion in FromPyObject

Lorak-mmk (changes requested):
Looks good, just one comment.

Lorak-mmk (approved):
The code you added looks fine. I did not read the ccm integration, because I plan for it to be rewritten later anyway. If it works its fine.

I do have some questions about code you did NOT add, because the PR is marked as fixing the issue.
- For stuff you added (HostFilter, AddressTranslator, TimestampGenerator) Rust Driver exposes sever builtin implementations, like: `SimpleTimestampGenerator`, `MonotonicTimestampGenerator`, `AcceptAllHostFilter`, `AllowListHostFilter`, `DcHostFilter`, `impl AddressTranslator for HashMap<SocketAddr, SocketAddr>`, `impl AddressTranslator for HashMap<&'static str, &'static str>`. We may want to expose some of them in Python instead of requiring users to implement everything. Builtin implementation could also be optimized by downcasting and thus using rust-based impl without going trough Python. It is fine to do this in another PR, if you open an issue.
- There are many more configuration parameters that should be trivial to implement, but I don't see them in this PR. Same as above, are you planning to do them in a follow-up? To name a few we would want: `auto_await_schema_agreement`, `hostname_resolution_timeout`, `compression`, `schema_agreement_interval`, `tcp_nodelay`, `tcp_keepalive_interval`, `use_keyspace`, `connection_timeout`, `disallow_shard_aware_port`, `local_ip_address`, `shard_aware_local_port_range`, `pool_size`, `metadata_request_serverside_timeout`, `keepalive_interval`, `keepalive_timeout`, `schema_agreement_timeout`, `auto_await_schema_agreement`, `refresh_metadata_on_auto_schema_agreement`, `write_coalescing_[delay]`, `custom_identity`

patrycja-ziemkiewicz:
I have asked you about bultins implementations and you told me to implement them later currently there is issue created to add them.

About methods not included I have PR with them prepared but I decided to split the code into smaller chunks for easier reviews.

Lorak-mmk:
> I have asked you about bultins implementations and you told me to implement them later currently there is issue created to add them.

Ok, I did not remember that. Sorry. I only remember talking about authenticator impls that are present in old Python Driver (but not in Rust Driver).

It is fine to postpose adding builtin impls and open an issue.

> About methods not included I have PR with them prepared but I decided to split the code into smaller chunks for easier reviews.

Great! That makes sense. In this case, could you change `Fixes: ` in PR description to `Ref: `?

Lorak-mmk:
@patrycja-ziemkiewicz I think this can be merged (after that please open issues for the next steps: one for builtin impls, other for more builder methods).

PR #62, opened 2026-04-07
Expose PreparedMetadata and ResultMetadata to Python

### LINKED ISSUE #61: Expose prepared and result metadata in Python API
The Rust driver already exposes metadata on `PreparedStatement`, but the Python-RS driver does not currently provide Python-facing access to it.

This issue is to expose:
- **prepared metadata**: metadata of a prepared statement about its bound values
- **result metadata**: metadata of a result set, including the number of columns and their specifications

Scope:
- add Python API on `PreparedStatement`
- expose metadata in a Python-friendly representation
- add typing stubs and minimal tests

pasinskik:
## Pre-review checklist

## Summary

This PR exposes query metadata from the Rust driver to Python.

It adds Python-facing metadata wrappers and exposes them through:

- `PreparedStatement.prepared_metadata`
- `PreparedStatement.result_metadata`
- `RequestResult.result_metadata`

## Motivation

The Rust driver already provides access to prepared-statement and result-set metadata, but the Python-RS driver did not expose it. This made it impossible for Python users to inspect bound-value metadata, result column metadata, or partition key index information even though the underlying driver already had it.

Exposing this metadata improves introspection, debugging, and overall API completeness.

## Changes

- add Python-facing metadata types:
  - `PyColumnSpec`
  - `PyPartitionKeyIndex`
  - `PyPreparedMetadata`
  - `PyResultMetadata`
- add `PreparedStatement.prepared_metadata`
- add `PreparedStatement.result_metadata`
- add `RequestResult.result_metadata`
- add typing stubs
- add integration tests for the new APIs

## Why this design

The goal of this PR is to expose query-related metadata without introducing a generic metadata abstraction that overlaps with other ongoing work.

For that reason:
- the new code is scoped specifically to query metadata
- Python-facing wrappers are introduced only for statement/result metadata
- metadata is exposed as properties, since it represents read-only object state rather than an action

This keeps the API small and Pythonic while still reflecting the structure of the underlying Rust driver.

Fixes: #61


src/query_metadata.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[pyo3(get)]
+    table_name: Option<String>,
+    /// The name of the keyspace containing the column.
+    /// This is `None` for columns that are not associated with a specific keyspace (e.g., result of an aggregation).
+    #[pyo3(get)]
+    keyspace_name: Option<String>,
+    /// The CQL type of the column.
+    #[pyo3(get)]
+    cql_type: String,
+}
```
patrycja-ziemkiewicz: You could use here `#[pyclass(get_all)]` instead of writing getters for all of this fields.

src/query_metadata.rs
```diff
@@ -0,0 +1,45 @@
...
+    let keyspace_name = table_spec.ks_name();
+
+    PyColumnSpec {
+        name: spec.name().to_owned(),
+        table_name: (!table_name.is_empty()).then(|| table_name.to_owned()),
+        keyspace_name: (!keyspace_name.is_empty()).then(|| keyspace_name.to_owned()),
+        // Temporarily use the debug representation of the CQL type.
+        cql_type: format!("{:?}", spec.typ()),
+    }
+}
```
patrycja-ziemkiewicz: You could implement trait `From<&ColumSpec> for PyColumnSpec`

src/query_metadata.rs
```diff
@@ -22,6 +23,90 @@ pub(crate) struct PyColumnSpec {
...
+        NativeType::Inet => "inet",
+        NativeType::SmallInt => "smallint",
+        NativeType::TinyInt => "tinyint",
+        NativeType::Time => "time",
+        NativeType::Timeuuid => "timeuuid",
+        NativeType::Uuid => "uuid",
+        NativeType::Varint => "varint",
+        _ => "unknown",
+    }
+}
```
patrycja-ziemkiewicz: I’m not sure about returning strings here. We could implement PyClasses to represent each type instead. This would be more memory-efficient and provide users with more information. In the current implementation, a user would have to parse a string to get the subtype of a list, which is quite inconvenient.

src/query_metadata.rs
```diff
@@ -22,6 +23,90 @@ pub(crate) struct PyColumnSpec {
...
+                .iter()
+                .map(column_type_to_cql)
+                .collect::<Vec<_>>()
+                .join(", ");
+            format!("tuple<{}>", inner)
+        }
+
+        _ => "unknown".to_owned(),
+    }
+}
```
patrycja-ziemkiewicz: Same here. We could definitely leave this implementation for the __repr__ method. However, I believe it would be more convenient for users to receive corresponding classes for other use cases.

src/query_metadata.rs
```diff
@@ -122,8 +134,20 @@ fn column_spec_to_py(spec: &scylla::frame::response::result::ColumnSpec<'_>) ->
...
+#[allow(dead_code)]
+fn partition_key_to_py(
+    pk: &scylla::frame::response::result::PartitionKeyIndex,
+) -> PyPartitionKeyIndex {
+    PyPartitionKeyIndex {
+        index: pk.index,
+        sequence_number: pk.sequence,
+    }
+}
+
```
patrycja-ziemkiewicz: You could implement From trait.

pasinskik: You're right, I will update it.

src/query_metadata.rs
pasinskik: I agree, we could implement PyClasses to mimic the structure from the Rust Driver and not stringify. It would be a better API, but I'm not sure if it fits in this PR or deserves its own. What do you think?

patrycja-ziemkiewicz: @wprzytula @Lorak-mmk what do you think?

Lorak-mmk: We should have typed info as you say.
It should be done in @jakub-graczyk Metadata PR, and this PR should resuse it.

src/query_metadata.rs
```diff
@@ -0,0 +1,170 @@
...
+    table_name: Option<String>,
+    /// The name of the keyspace containing the column.
+    /// This is `None` for columns that are not associated with a specific keyspace (e.g., result of an aggregation).
+    #[pyo3(get)]
+    keyspace_name: Option<String>,
+    /// The CQL type of the column.
+    #[pyo3(get)]
+    cql_type: Py<PyCqlColumnType>,
+}
+
```
patrycja-ziemkiewicz: You can write here ```#[pyclass(get_all)]``` instead of having ```#[pyo3(get)]``` under each attribute.

src/query_metadata.rs
```diff
@@ -0,0 +1,170 @@
...
+#[pyclass(name = "ResultMetadata", skip_from_py_object, frozen)]
+#[derive(Clone)]
+pub(crate) struct PyResultMetadata {
+    /// The number of columns in the query result.
+    #[pyo3(get)]
+    column_count: usize,
+    /// The specifications of the columns in the query result.
+    #[pyo3(get)]
+    columns: Vec<PyColumnSpec>,
+}
```
patrycja-ziemkiewicz: You can write here #[pyclass(get_all)] instead of having #[pyo3(get)] under each attribute.

src/query_metadata.rs
```diff
@@ -0,0 +1,170 @@
...
+    let cql_type = extract_column_type(py, spec.typ())
+        .map_err(DriverQueryMetadataError::column_type_extraction_failed)?;
+
+    Ok(PyColumnSpec {
+        name: spec.name().to_owned(),
+        table_name: (!table_name.is_empty()).then(|| table_name.to_owned()),
+        keyspace_name: (!keyspace_name.is_empty()).then(|| keyspace_name.to_owned()),
+        cql_type,
+    })
+}
```
patrycja-ziemkiewicz: It would be better to implement ```From``` trait here.

src/query_metadata.rs
patrycja-ziemkiewicz: This conversation was resolved but I do not see this changes implemented. I still believe this should be implemented using From trait.

src/query_metadata.rs
```diff
@@ -0,0 +1,170 @@
...
+    #[pyo3(get)]
+    name: String,
+    /// The name of the table containing the column.
+    /// This is `None` for columns that are not associated with a specific table (e.g., result of an aggregation).
+    #[pyo3(get)]
+    table_name: Option<String>,
+    /// The name of the keyspace containing the column.
+    /// This is `None` for columns that are not associated with a specific keyspace (e.g., result of an aggregation).
+    #[pyo3(get)]
+    keyspace_name: Option<String>,
```
patrycja-ziemkiewicz: Could you explain why table_name and keyspace_name can be None? As I do not really get this comment. In rust driver tableSpec always return &str.

src/query_metadata.rs
pasinskik: Thanks for re-reviewing this. Just to clarify: this PR was blocked for a while by the metadata PR, and instead of rebasing the old version I ended up rewriting the changes on top of the current code. Because of that, some previous review threads were accidentally resolved even though the same concerns still applied to the rewritten code. Sorry for the noise; I’ll go through the repeated comments and address them properly once again

src/query_metadata.rs
pasinskik: I thought that the `&str` could be empty, e.g. for `SELECT count(*) FROM table`, but I tested it and it seems that this is not the case here. I will adjust the code.

src/query_metadata.rs
pasinskik: I have updated it to use `TryFrom<>` (it can fail during column type extraction).

python/scylla/_rust/errors.pyi
Lorak-mmk, line 1: Commit: "Add query_metadata module and PyColumnSpec"

`query` is not a name that should be used here. Query can refer to either:
- QUERY message from binary CQL protocol
- SELECT request

`ColumnSpec` here is about neither of those.

I'm not sure what is a good name here. Or if the module is even needed - perhaps its contents could be distributed to other modules, if each fits somewhere.

src/query_metadata.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// The name of the column.
+    name: String,
+    /// The name of the table containing the column.
+    table_name: String,
+    /// The name of the keyspace containing the column.
+    keyspace_name: String,
+    /// The CQL type of the column.
+    cql_type: Py<PyCqlColumnType>,
+}
+
```
Lorak-mmk, line 19: Commit: "Add query_metadata module and PyColumnSpec"

The Python-side definition is:
```python
class ColumnSpec:
    """
    Specification of a column in a result set, used for both prepared statement metadata and query result metadata.
    """

    @property
    def name(self) -> str: ...
    @property
    def table_name(self) -> str: ...
    @property
    def keyspace_name(self) -> str: ...
    @property
    def cql_type(self) -> CqlColumnType: ...
```

How does that even work?
Afaik Pyo3 has no access to our type stubs when generating its code, and I'm pretty sure it doesn't generate accessors for all Rust fields. So how Python code accessing those properties work?

Also, you are storing them as Rust String. Regardless of how they are access, it will require costly conversion to PyString on each access.

src/query_metadata.rs
```diff
@@ -40,8 +40,29 @@ impl<'a> TryFrom<(Python<'a>, &scylla::frame::response::result::ColumnSpec<'a>)>
...
+/// Specification of a partition key index in prepared statement metadata.
+#[pyclass(name = "PartitionKeyIndex", skip_from_py_object, frozen, get_all)]
+#[derive(Clone)]
+pub(crate) struct PyPartitionKeyIndex {
+    /// The index of the partition key.
+    index: u16,
+    /// The sequence number of the partition key, used for multi-column partition keys.
+    sequence_number: u16,
+}
+
```
Lorak-mmk, line 52: ditto about access and conversions

python/scylla/_rust/query_metadata.pyi
```diff
@@ -25,3 +25,13 @@ class PartitionKeyIndex:
...
+
+class PreparedMetadata:
+    """
+    Metadata for a prepared statement, including column specifications and partition key indexes.
+    """
+
+    @property
+    def columns(self) -> list[ColumnSpec]: ...
+    @property
+    def partition_key_indexes(self) -> list[PartitionKeyIndex]: ...
```
Lorak-mmk, line 37: And here as well.

src/query_metadata.rs
```diff
@@ -100,10 +100,77 @@ impl<'a> TryFrom<(Python<'a>, &scylla::statement::prepared::PreparedStatement)>
...
+            .iter()
+            .map(|spec| PyColumnSpec::try_from((py, spec)))
+            .collect::<Result<Vec<PyColumnSpec>, DriverQueryMetadataError>>()?;
+
+        Ok(PyResultMetadata {
+            column_count: columns.len(),
+            columns,
+        })
+    }
+}
```
Lorak-mmk, line 167: Having to define TryFrom for tuples is imo a clear sign that TryFrom is not a good fit here. Those should be constructors on relevant types.

src/query_metadata.rs
```diff
@@ -60,9 +60,50 @@ impl From<&scylla::frame::response::result::PartitionKeyIndex> for PyPartitionKe
...
+            .map(PyPartitionKeyIndex::from)
+            .collect();
+
+        Ok(PyPreparedMetadata {
+            columns,
+            partition_key_indexes,
+        })
+    }
+}
+
```
Lorak-mmk, line 102: ditto

src/statement.rs
```diff
@@ -139,6 +140,22 @@ impl PyPreparedStatement {
...
+    ) -> Result<PyPreparedMetadata, DriverQueryMetadataError> {
+        PyPreparedMetadata::try_from((py, &self._inner))
+    }
+
+    #[getter]
+    fn get_result_metadata(
+        &self,
+        py: Python<'_>,
+    ) -> Result<PyResultMetadata, DriverQueryMetadataError> {
+        PyResultMetadata::try_from((py, &self._inner))
```
Lorak-mmk, line 157: Commit: "Add query metadata API for PreparedStatement"

First, and observation: here the Python properties have Rust-side getters. I can imagine that Pyo3 matches the getter with the attribute name, after stripping `get_` prefix. I still cannot imagine how property access works on other types you introduced.

Second, a problem: each access will create a new `PyPreparedMetadata` / `PyResultMetadata` object. This should be lazily constructed and cached.

src/deserialize/results.rs
```diff
@@ -200,6 +204,15 @@ impl RequestResult {
...
     }
+
+    /// Returns metadata about the query result.
+    #[getter]
+    fn get_result_metadata(
+        &self,
+        py: Python<'_>,
+    ) -> Result<PyResultMetadata, DriverQueryMetadataError> {
+        PyResultMetadata::try_from((py, self.query_result.as_ref()))
+    }
```
Lorak-mmk, line 215: ditto about lazy computation.

src/statement.rs
Lorak-mmk: Warning: need to be careful about caching result metadata on PreparedStatement, because it may change. Need to detect that and recalculate when necessary.

Lorak-mmk (changes requested):

src/query_metadata.rs
patrycja-ziemkiewicz: He is using get_all pyo3 macro that should be used when we want to expose all struct fields as properties to python https://docs.rs/pyo3/latest/pyo3/attr.pyclass.html.

Lorak-mmk: Ah, I see.

Let's not use it. It may hide costly conversions.

patrycja-ziemkiewicz: I thought `get_all` was acceptable here because it reduces boilerplate when the whole class is exposed to Python anyway. In that case, the conversions are still fairly visible to me, since using `get_all` means we are deliberately exposing all fields of the class.

But I agree explicit getters make the conversion cost more visible. If we prefer that style, I can open an issue to remove existing `get_all` usages and add a lint/check to prevent new ones.

patrycja-ziemkiewicz:
https://github.com/scylladb/python-rs-driver/pull/169 replaced by this PR

PR #65, opened 2026-04-19
Two-phase lazy cache with freeze semantics

jakub-graczyk:
Adds a `Cache<K, V>` supporting lazy initialization and immutable snapshot freezing for exposing metadata.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -68,6 +68,9 @@ jobs:
                   echo "Waiting for nodes... (attempt $i/30)" && sleep 1 ||
                   { echo "All nodes ready"; break; }; done
 
+            - name: Run native Rust tests
+              run: cargo test --no-default-features
+
             - name: Run tests with Python 3.10
               run: uv run --python 3.10 pytest -v
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,8 @@
 use std::sync::LazyLock;
 
+#[cfg(test)]
+mod tests;
+
 use crate::deserialize::value;
 use deserialize::results;
 use pyo3::prelude::*;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,7 @@ use pyo3::prelude::*;
 use tokio::runtime::Runtime;
 
 mod batch;
+mod cache;
 mod deserialize;
 mod enums;
 mod errors;
--- a/src/tests/cache_tests.rs
+++ b/src/tests/cache_tests.rs
@@ -1,0 +1,248 @@
+use pyo3::prelude::*;
+use pyo3::types::PyInt;
+use std::sync::{
+    Arc,
+    atomic::{AtomicUsize, Ordering},
+};
+
+use crate::cache::Cache;
+
+#[test]
+fn get_or_init_inserts_and_reuses() {
+    Python::initialize();
+    Python::attach(|py| {
+        let cache = Cache::<String, PyInt>::new();
+        let calls = Arc::new(AtomicUsize::new(0));
+
+        let key = "a".to_string();
+
+        let calls1 = calls.clone();
+        let v1 = cache
+            .get_or_init(py, &key, |_| {
+                calls1.fetch_add(1, Ordering::SeqCst);
+                Ok(Some(42i64.into_pyobject(py)?.unbind()))
+            })
+            .unwrap()
+            .unwrap();
+
+        let calls2 = calls.clone();
+        let v2 = cache
+            .get_or_init(py, &key, |_| {
+                calls2.fetch_add(1, Ordering::SeqCst);
+                Ok(Some(999i64.into_pyobject(py)?.unbind()))
+            })
+            .unwrap()
+            .unwrap();
+
+        assert_eq!(calls.load(Ordering::SeqCst), 1);
+
+        let ptr1 = v1.bind(py).as_ptr();
+        let ptr2 = v2.bind(py).as_ptr();
+        assert!(ptr1.eq(&ptr2));
+    });
+}
+
+#[test]
+fn get_or_init_none_does_not_insert() {
+    Python::initialize();
+    Python::attach(|py| {
+        let cache = Cache::<String, PyInt>::new();
+        let key = "missing".to_string();
+
+        let r = cache.get_or_init(py, &key, |_| Ok(None)).unwrap();
+        assert!(r.is_none());
+
+        let r2 = cache.get_or_init(py, &key, |_| Ok(None)).unwrap();
+        assert!(r2.is_none());
+    });
+}
+
+#[test]
+fn freeze_blocks_new_keys_and_exposes_mapping() {
+    Python::initialize();
+    Python::attach(|py| {
+        let cache = Cache::<String, PyInt>::new();
+
+        // Pre-populate one key in active mode.
+        let k1 = "x".to_string();
+        cache
+            .get_or_init(py, &k1, |_| Ok(Some(1i64.into_pyobject(py)?.unbind())))
+            .unwrap();
+
+        // Freeze with additional entries.
+        let proxy = cache
+            .get_or_init_python_mapping(py, || {
+                vec![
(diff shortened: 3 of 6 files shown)


wprzytula:
Waiting for unit tests before the review.
Also, please try to avoid allocations for string lookup, as discussed in the call.

src/lib.rs
```diff
@@ -21,6 +31,285 @@ use crate::utils::add_submodule;
...
+/// ## 1. Active phase
+/// - The cache behaves as a concurrent lazy-initialized map
+/// - Keys may be inserted on-demand via `get_or_init`
+/// - Missing values are computed using `f` closure
+/// - Computation is intentionally performed outside locks to reduce contention
+///
+/// ## 2. Frozen phase
+/// - The cache becomes immutable in terms of keyspace
+/// - No new keys may be inserted
+/// - A stable Python `MappingProxy` view is created and returned
```
Lorak-mmk: If this is supposed to be a generic cache struct, then `keyspace` should not appear in this description.

src/lib.rs
```diff
@@ -21,6 +31,285 @@ use crate::utils::add_submodule;
...
+    K: Eq + Hash + Clone + Sync + Send,
+{
+    pub fn new() -> Self {
+        Self {
+            state: RwLock::new(CacheState::Active {
+                map: HashMap::new(),
+            }),
+        }
+    }
+
```
Lorak-mmk: In general you should put restrictions on impl blocks, not on structs. Putting them on structs forces you to repeat them absolutetly everywhere, and doesn't have much advantages in this case.
For example, look at std hashmap, it has no restrictions on struct definition:
https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#247-254
`new()` also doesn't have restrictions: https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#256-273

The restrictions only appear when they are actually necessary, like in `get()`: https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#997-1003 which is put in an impl block with restrictions: https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#852-857

src/lib.rs
```diff
@@ -21,6 +31,285 @@ use crate::utils::add_submodule;
...
+            if matches!(&*read_guard, CacheState::Frozen { .. }) {
+                return Ok(None);
+            }
+        }
+
+        // Avoid work inside write lock.
+        // Tradeoff wasted work for lock contention.
+        let Some(resolved) = f(key)? else {
+            return Ok(None);
+        };
```
Lorak-mmk: Why does `F` return Option?

src/lib.rs
```diff
@@ -21,6 +31,285 @@ use crate::utils::add_submodule;
...
+            }
+        }
+
+        // Avoid work inside write lock.
+        // Tradeoff wasted work for lock contention.
+        let entries: Vec<(K, Py<V>)> = f()
+            .into_iter()
+            .map(|(k, v_res)| Ok((k, v_res?)))
+            .collect::<PyResult<Vec<_>>>()?;
+
```
Lorak-mmk: In the real code, what will `f` be? I don't get why it returns `Vec<(K, PyResult<Py<V>>)>` instead of `PyResult<Vec<(K, Py<V>)>>`

Lorak-mmk (changes requested):
Looks fine I think.

src/lib.rs
jakub-graczyk: It will allocate a Vec of `Py<T>`.  Creating a new `Py<T>` is fallible so for convenience I convert it here to `PyResult<Vec<(K, Py<V>)>>`  and propagate the error instead of doing it in every closure.

src/lib.rs
jakub-graczyk: For example `get_keyspace` can return None when keyspace doesn't exist. In such case we will call `f` on each call to let's say `get_keyspace("non_existent_ks")` which (`f`) will return `None` and cache will propagate the cache miss with `Ok(None)`. I believe there's no point in storing `Option` in the cache as `f` will not allocate `Py<T>` in the `None` case and it's outside the lock.

src/lib.rs
jakub-graczyk: I'm not referring to Scylla's keyspace, I meant the hashmap's key space as in domain of keys. Will change to "key domain"

Lorak-mmk (approved):

src/lib.rs
wprzytula, line 1: ❓ Do new functions introduced here need to be `pub`?

wprzytula, line 1: ♻️ I believe the cache deserves its own module file, e.g., `cache.rs`.

.github/workflows/CI.yml
```diff
@@ -68,6 +68,9 @@ jobs:
                   echo "Waiting for nodes... (attempt $i/30)" && sleep 1 ||
                   { echo "All nodes ready"; break; }; done
 
+            - name: Run native Rust tests
+              run: cargo test --no-default-features
```
wprzytula, line 72: ⚠️  Running this locally results in:
`rust-lld: error: unable to find library -lpython3.13`

wprzytula: Even though I have python3 installed:
```
❯ python3
Python 3.13.7 (main, Mar  3 2026, 12:19:54) [GCC 15.2.0] on linux
```

src/tests/cache_tests.rs
```diff
@@ -0,0 +1,251 @@
...
+                .unwrap()
+                .unwrap();
+
+            assert_eq!(calls.load(Ordering::SeqCst), 1);
+
+            let ptr1 = v1.bind(py).as_ptr();
+            let ptr2 = v2.bind(py).as_ptr();
+            assert!(ptr1.eq(&ptr2));
+        });
+    }
```
wprzytula: ⛏️ Let's use `&str` as a key, to prove that borrowed types work, too.

src/lib.rs
```diff
@@ -22,6 +32,272 @@ use crate::utils::add_submodule;
...
+                return Ok(view.clone_ref(py).into_bound(py));
+            }
+        }
+
+        // Avoid work inside write lock.
+        // Tradeoff wasted work for lock contention.
+        let entries: Vec<(K, Py<V>)> = f()
+            .into_iter()
+            .map(|(k, v_res)| Ok((k, v_res?)))
+            .collect::<PyResult<Vec<_>>>()?;
```
wprzytula: 💭 Shouldn't we add some context to the error here before propagating it?

wprzytula (changes requested):
Opened https://github.com/jakub-graczyk/python-rs-driver/pull/1 with two improvements.

.github/workflows/CI.yml
Lorak-mmk: You need python3-dev or something like that

src/tests/cache_tests.rs
jakub-graczyk: There are tests for &str below.

src/lib.rs
jakub-graczyk: We are converting errors only when returning to Python. This is a reusable struct so it can’t have typed errors. This is basically candidate for unwrap() as error when creating a Py<T> is unlikely (even pyo3 panics when PyString can’t be created due to oom and so on for convenience). I believe no additional context is required for this error.

.github/workflows/CI.yml
wprzytula: This helped, thanks.

src/lib.rs
wprzytula: OK

src/lib.rs
jakub-graczyk: It’s a Python library so it doesn’t really matter, but let’s use pub(crate)

wprzytula: Why do you think it doesn't matter?
`pub` functions can't be fully optimised out, because they must be linkable against.
This is similar to a C function that misses the `static` keyword.

Lorak-mmk (approved):

Issue #64, opened 2026-04-14
Use Pyo3 extension traits for blocking methods to avoid deadlock with GIL


jakub-graczyk:
While performing blocking operations on Mutex an extension method `lock_py_attached()` defined in [extension trait](https://pyo3.rs/main/doc/pyo3/sync/trait.mutexext) should be used to avoid deadlocks with GIL and conflict with Python's GC.

Similar methods from respective extension traits should be used when using any other synchronization mechanisms from standard library. 

We would like to replace all unsafe methods used in the project in addition to configuring `clippy::disallowed_methods` in `config.toml` to disallow blocking methods that don't acknowledge holding the GIL.
This way using potentially dangerous methods will require explicit allow lint.

wprzytula:
Same goes for OnceLock - see #58.

PR #66, opened 2026-04-20
Integrate Rust logging with Python logging

### LINKED ISSUE #52: Integrate Rust logging with Python logging
Currently, Rust emits logs using `tracing` crate, using the [`Formatter` Subscriber](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/type.Formatter.html) found in `tracing_subscriber`. It emits logs to stdout, and is not controllable by Python logging framework.  We'd like to integrate both logging frameworks, so that Python controls log level, among others.

paulinaczajkowska:
Fixes: #52 

## Description

Enabling `tracing` with the `log` feature causes trace instrumentation points to emit log records as well as trace events, if a default tracing subscriber has not been set. As a result, `pyo3-log` can capture these logs and forward them into Python’s standard logging system.

Logging verbosity can be controlled using standard Python logging levels: CRITICAL (50), ERROR (40), WARNING (30), INFO (20), DEBUG (10).

In addition, a more detailed TRACE level (more verbose than DEBUG) is supported by using level `5`:

```python
logging.basicConfig(level=5, ...)
```

## Pre-review checklist

- [ ] ~~I have provided docstrings for the public items that I want to introduce.~~
- [ ] ~~I have adjusted the documentation in `./docs/source/`.~~

### DIFF
--- a/python/tests/test_logging.py
+++ b/python/tests/test_logging.py
@@ -1,0 +1,26 @@
+import logging
+
+import pytest
+from _pytest.logging import LogCaptureFixture
+from scylla.session_builder import SessionBuilder
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_rust_logs_forwarded(caplog: LogCaptureFixture):
+    logging.basicConfig(level=logging.INFO)
+    caplog.set_level(logging.INFO)
+
+    session = await SessionBuilder().contact_points([("127.0.0.2", 9042)]).connect()
+
+    # Setting replication factor to 1 generates the following warning log:
+    # Response from the database contains a warning warning="Using
+    # Replication Factor replication_factor=1 lower than the
+    # minimum_replication_factor_warn_threshold=3 is not recommended."
+    await session.execute(
+        "CREATE KEYSPACE IF NOT EXISTS example_ks WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1};"
+    )
+    await session.execute("USE example_ks;")
+
+    # Assert that the warning log was captured
+    assert any("Response from the database contains a warning" in rec.getMessage() for rec in caplog.records)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,8 @@ mod tests;
 use crate::deserialize::value;
 use deserialize::results;
 use pyo3::prelude::*;
+use pyo3::sync::OnceExt;
+use std::sync::Once;
 use tokio::runtime::Runtime;
 
 mod batch;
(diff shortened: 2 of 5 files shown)


python/tests/test_logging.py
```diff
@@ -0,0 +1,23 @@
...
+    host = "127.0.0.2"
+    port = 9042
+    session = await SessionBuilder([host], port).connect()
+
+    await session.execute(
+        "CREATE KEYSPACE IF NOT EXISTS example_ks WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1};"
+    )
+    await session.execute("USE example_ks;")
+
+    assert any("Response from the database contains a warning" in rec.getMessage() for rec in caplog.records)
```
jakub-graczyk, line 26: Why are assert against a message about a warning?, looks flaky.

src/lib.rs
```diff
@@ -21,10 +22,22 @@ use crate::utils::add_submodule;
...
+            eprintln!("pyo3_log::try_init failed: {:?}", e)
+        }
+    });
+}
+
 /// A Python module implemented in Rust.
 #[pymodule]
 #[pyo3(name = "_rust")]
 fn scylla(py: Python<'_>, module: &Bound<'_, PyModule>) -> PyResult<()> {
+    init_logging();
```
jakub-graczyk: Why do we wrap it in Once?
```suggestion
    pyo3_log::init();
```

paulinaczajkowska: Logger can be installed only once. In case a logger is already installed, an error is returned by `pyo3_log::init()`, so `Once` is here to avoid getting this error.

jakub-graczyk: Will the logger be initialized in any other place than the main `scylla` module?

Lorak-mmk: Most likely not. More important question: is it possible for `scylla` to be called more than once?

Lorak-mmk (commented):
Was the solution implemented in this PR (using `tracing` with `log` feature and `pyo3-log`) difficult to figure out? I'm wondering why it took over a month.

src/lib.rs
```diff
@@ -22,11 +23,22 @@ use crate::utils::add_submodule;
 
 pub static RUNTIME: LazyLock<Runtime> = LazyLock::new(|| Runtime::new().unwrap());
 
+static INIT_LOG: Once = Once::new();
+
+fn init_logging() {
+    INIT_LOG.call_once(|| {
```
jakub-graczyk: Regardless of whether it should be wrapped in `Once` please look at #64 and this trait for `Once` https://pyo3.rs/main/doc/pyo3/sync/trait.onceext

Lorak-mmk (approved):

jakub-graczyk (approved):

wprzytula (approved):

PR #67, opened 2026-04-22
Session builder v2

patrycja-ziemkiewicz:
Added Session builder methods

Ref: https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/6

## Pre-review checklist


src/session_builder.rs
```diff
@@ -136,6 +148,30 @@ impl SessionBuilder {
     }
 }
 
+
+struct PyDuration(Duration);
+
+impl<'py> FromPyObject<'_, 'py> for PyDuration {
```
jakub-graczyk, line 631: Checkout https://pyo3.rs/main/doc/pyo3/time/ there's already timedelta to Duration conversion in pyo3

jakub-graczyk (changes requested):
Please look at ExecutionProfile's stub file, let's provide default values for session builder method signatures and remove information about default value from docstring.

patrycja-ziemkiewicz:
> Please look at ExecutionProfile's stub file, let's provide default values for session builder method signatures and remove information about default value from docstring.

These methods exist to change defaults, not define them. Adding default values to the signature is misleading. It suggests the argument is optional, while in reality the whole point is to explicitly override something.

python/tests/test_session_builder.py
```diff
@@ -347,3 +347,18 @@ async def test_custom_host_filter_fallback_on_failure(
...
+        "127.0.0.1",
+        "::1",
+        ipaddress.IPv4Address("127.0.0.2"),
+        ipaddress.IPv6Address("::2"),
+        None,
+    ],
+)
+def test_local_ip_address_valid_formats(ip: Any):
+    builder = SessionBuilder().local_ip_address(ip)
+    assert isinstance(builder, SessionBuilder)
```
Lorak-mmk: This is not a very comprehensive test. What you could do is bind to some address like `127.0.0.123`, call `ss` command and verify that there is some socket bound to this address. AI should handle that easily and quickly.

python/scylla/_rust/session_builder.pyi
```diff
@@ -181,3 +181,29 @@ class SessionBuilder:
...
+        A possible use case is when you want to have multiple [`Session`] objects and do not want
+        them to compete for the ports within the same range. It is then advised to assign
+        mutually non-overlapping port ranges to each session object.
+
+        By default the driver uses port range ``(49152, 65535)``.
+
+        **Validation Rules:**
+        A ``SessionConfigError`` is raised if:
+        1. The range is empty (``end`` < ``start``).
+        2. The range starts below port ``1024`` (reserved system ports).
```
Lorak-mmk, line 272: In Rust it is obvious that the range is right-closed because of the types used. Here it is less obvious, it would be good to mention that explicitly.

src/session_builder.rs
```diff
@@ -136,6 +147,26 @@ impl SessionBuilder {
...
+        if let Ok(secs) = obj.extract::<f64>() {
+            let duration = Duration::try_from_secs_f64(secs)
+                .map_err(|_| DriverSessionConfigError::invalid_duration(obj))?;
+            return Ok(PyDuration(duration));
+        }
+
+        Err(DriverSessionConfigError::invalid_duration(obj))
+    }
+}
+
```
Lorak-mmk: Isn't there some builtin support for Duration in Pyo3?

python/scylla/_rust/session_builder.pyi
```diff
@@ -566,3 +566,30 @@ class SessionBuilder:
...
+
+        Parameters
+        ----------
+        enable : bool
+
+        Returns
+        -------
+        SessionBuilder
+        """
+        ...
```
Lorak-mmk, line 642: In Rust Driver we have 2 separate settings: one with bool, another (`write_coalescing_delay`) that configures the delay. This is because delay was added later.
Here we can avoid this, and have 1 options that accepts an optional value, instead of having 2 separate configs.

python/scylla/_rust/enums.pyi
```diff
@@ -57,3 +57,33 @@ class PoolSize:
...
+        Parameters
+        ----------
+        delay : int
+            Delay in milliseconds. Must be greater than 0.
+
+        Returns
+        -------
+        WriteCoalescingDelay
+        """
+        ...
```
Lorak-mmk, line 110: Isn't the convention in Python to pass time using seconds as `float`? We need to have one way to pass durations around and stick to it. What do we do in other parts of the code?

src/enums.rs
```diff
@@ -139,12 +140,135 @@ impl PyWriteCoalescingDelay {
...
+            inner.set_application_name(v);
+        }
+        if let Some(v) = application_version {
+            inner.set_application_version(v);
+        }
+        if let Some(v) = client_id {
+            inner.set_client_id(v);
+        }
+
+        Self { inner }
```
Lorak-mmk, line 281: There is a problem: we want the driver to advertise itself as Python-rs Driver by default, not as Rust Driver.
So you need to, before assigning to the field in SessionConfig (`slf.config.identity = identity.inner;`), check if user set their custom name / version. If not, set them to defaults correct for this driver.

Why this way? Because we want to:
- Have correct default
- Have `custom_driver_name` / `custom_driver_version` getters only return non-None if the user set the values there.

Lorak-mmk (changes requested):

src/session_builder.rs
patrycja-ziemkiewicz: Yes there is but only for timedelta types. We want to also enable giving floats which would be treated as number of seconds.

src/session_builder.rs
```diff
@@ -97,6 +102,217 @@ impl SessionBuilder {
...
+        Ok(slf)
+    }
+
+    fn keepalive_timeout(mut slf: PyRefMut<'_, Self>, timeout: PyDuration) -> PyRefMut<'_, Self> {
+        if timeout.0 <= Duration::from_secs(1) {
+            log::warn!(
+                "Setting the keepalive timeout to low values ({:?}) is not recommended as it can have a negative impact on performance. Consider setting it above 5 second.",
+                timeout.0
+            );
+        }
```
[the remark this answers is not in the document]
patrycja-ziemkiewicz: @Lorak-mmk @wprzytula I copied this warning message from the rust driver. What is the desired behaviour here?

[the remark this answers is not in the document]
wprzytula: We explicitly warn against very small keepalice timeouts (<=1s) and advise significantly more (>=5s). It's all OK.

src/session_builder.rs
```diff
@@ -111,6 +111,27 @@ impl SessionBuilder {
...
+        if let Ok(secs) = obj.extract::<f64>() {
+            let duration = Duration::try_from_secs_f64(secs)
+                .map_err(|_| DriverSessionConfigError::invalid_duration(obj))?;
+            return Ok(PyDuration(duration));
+        }
+
+        Err(DriverSessionConfigError::invalid_duration(obj))
+    }
+}
+
```
Lorak-mmk, line 647: Commit: "Refractor ContactPoints"

Why is this change present in this commit? I don't see how PyDuration is related to contact points.

python/scylla/_rust/session_builder.pyi
```diff
@@ -2,13 +2,21 @@ from __future__ import annotations
...
+class SessionBuilderConfig:
+    """
+    A read-only snapshot of the current SessionBuilder configuration.
+
+    This object is frozen and represents the immutable state of the builder
+    at the moment it was generated.
+    """
 class SessionBuilder:
     """
     Builder for configuring and creating a :class:`Session`.
```
Lorak-mmk, line 86: Why do we need both SessionBuilder and SessionBuilderConfig? It is not explained in the commit message at all (in fact the commit message is just "Add SessionBuilderConfig").

I get why we can't store Rust Drivers SessionConfig underneath and just get values from it - there is no good way to convert some to Python repr, especially if we don't want to recreate object on each get.

I don't understand why we can't have only `SessionBuilder` storing all the data in the way newly introduced `SessionBuilderConfig` does, allowing both setting and getting. They would be transfered to Rust Driver SessionBuilder when constructing Session.

src/session_builder.rs
```diff
@@ -8,99 +8,143 @@ use crate::policies::{
 };
```
Lorak-mmk, line 9: Commit: "Refractor SessionBuilder"

Why is there no commit message? what do you want to achieve with the refactor, what was wrong before?

python/scylla/_rust/enums.pyi
```diff
@@ -20,3 +23,88 @@ class SerialConsistency(IntEnum):
...
+        Parameters
+        ----------
+        delay : float | timedelta
+            The delay duration. If float, it represents seconds.
+
+        Returns
+        -------
+        WriteCoalescingDelay
+        """
+        ...
```
Lorak-mmk, line 110: Commit: "Add pool_size to SessionBuilder"

I don't think this belongs in this commit

python/scylla/_rust/session_builder.pyi
```diff
@@ -414,3 +414,21 @@ class SessionBuilder:
...
+        ----------
+        interval : timedelta | float
+            The interval duration. If a ``float`` is provided,
+            it is interpreted as **seconds**.
+
+        Returns
+        -------
+        SessionBuilder
+        """
+        ...
```
Lorak-mmk: This mimics the API from Rust Driver, which has a problem. It is not possible to disable the keepalive without modifying `SessionConfig`. Here it is not possible to disable it at all. `None` should be accepted as well.

python/scylla/_rust/session_builder.pyi
```diff
@@ -432,3 +432,21 @@ class SessionBuilder:
...
+        ----------
+        timeout : timedelta | float
+            The timeout duration. If a ``float`` is provided,
+            it is interpreted as **seconds**.
+
+        Returns
+        -------
+        SessionBuilder
+        """
+        ...
```
Lorak-mmk, line 536: ditto

python/scylla/_rust/session_builder.pyi
```diff
@@ -604,3 +604,21 @@ class SessionBuilder:
...
+        ----------
+        interval : timedelta | float
+            The interval duration. If a ``float`` is provided,
+            it is interpreted as **seconds**.
+
+        Returns
+        -------
+        SessionBuilder
+        """
+        ...
```
Lorak-mmk, line 632: Can we skip that for now? We will be changing how metadata is fetched in Rust Driver, and it may also impact how fetching is configured. I don't want to introduce an option that may soon be deprecated in Rust Driver.

Lorak-mmk (changes requested):

src/session_builder.rs
patrycja-ziemkiewicz: I staged it accidentally. I will move it to commit in which it should be introduced.

src/session_builder.rs
patrycja-ziemkiewicz: The refactor fixes a thread-safety issue: previously, we used PyRefMut for mutability, which relies on a RefCell under the hood rather than a Mutex. This change adds explicit mutexes to the SessionBuilder internal fields and freezes the class, ensuring it is fully thread-safe. I will add this to commit description.

python/scylla/_rust/session_builder.pyi
patrycja-ziemkiewicz: I introduced SessionBuilderConfig so we could have both the underlying Rust driver config and the Python-specific conversion fields under a single Mutex, which is why keeping it as a distinct internal struct is necessary.

Regarding the user-facing API: I made a decision to provide a read-only config snapshot via config() method. However, if you think it would be better to have getters and setters directly on SessionBuilder, I can absolutely switch to that. Just note that we would need to rename the  builder methods (e.g., adding a with_ prefix) since Python doesn't allow a property and a method to share the exact same name.

Lorak-mmk: > we would need to rename the builder methods (e.g., adding a with_ prefix) since Python doesn't allow a property and a method to share the exact same name.

Ah, that makes sense. Let's keep what we have now.

src/session_builder.rs
```diff
@@ -111,6 +113,52 @@ impl SessionBuilder {
...
+            execution_profile,
+            shard_aware_local_port_range: (49152, 65535),
+            contact_points: Vec::new(),
+            host_filter: None,
+            authenticator: None,
+            address_translator: None,
+            timestamp_generator: None,
+        })
+    }
+}
```
Lorak-mmk: Why do we need a Python-side constructor for this? Won't it be created only as the snapshot of current builder?

src/session_builder.rs
```diff
@@ -427,6 +427,32 @@ impl SessionBuilder {
...
+    ) -> PyRefMut<'_, Self> {
+        if let Some(delay) = delay {
+            slf.config.write_coalescing_delay = delay.inner;
+            slf.config.enable_write_coalescing = true;
+        } else {
+            slf.config.enable_write_coalescing = false;
+        }
+        slf
+    }
+
```
Lorak-mmk, line 439: @wprzytula I don't think we have capability to fetch tracing info yet in python-rs.
Do you think it makes any sense to add tracing-related SessionBuilder options now, or maybe we should add them when we add tracing fetching?

wprzytula: At most we can write dead code for now (meaning that it's hidden from a user).
I wouldn't expose to a user config that is doing nothing.

Lorak-mmk: @patrycja-ziemkiewicz Let's remove tracing-related stuff, rebase on main, and then we can merge I think.

patrycja-ziemkiewicz: Done

src/enums.rs
```diff
@@ -78,9 +85,249 @@ impl From<SerialConsistency> for PySerialConsistency {
...
+    }
+
+    #[getter]
+    fn kind<'py>(&self, py: Python<'py>) -> &Bound<'py, PyString> {
+        match self.inner {
+            WriteCoalescingDelay::SmallNondeterministic => intern!(py, "small_nondeterministic"),
+            WriteCoalescingDelay::Milliseconds(_) => intern!(py, "milliseconds"),
+            _ => unimplemented!(),
+        }
+    }
```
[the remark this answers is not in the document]
Lorak-mmk: @patrycja-ziemkiewicz Copilot found a real problem, but suggested a wrong solution.
You should put `#[deny(clippy::wildcard_enum_match_arm)]` on such matches.

src/enums.rs
```diff
@@ -78,9 +85,249 @@ impl From<SerialConsistency> for PySerialConsistency {
...
+            WriteCoalescingDelay::Milliseconds(ms) => PyString::from_fmt(
+                py,
+                format_args!(
+                    "WriteCoalescingDelay(kind={}, milliseconds={})",
+                    self.kind(py),
+                    ms.get()
+                ),
+            )?,
+            _ => unimplemented!(),
+        };
```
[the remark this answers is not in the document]
Lorak-mmk: ditto about `#[deny(clippy::wildcard_enum_match_arm)]`

src/session_builder.rs
```diff
@@ -8,97 +9,440 @@ use crate::policies::{
...
+        py: Python<'py>,
         contact_points: ContactPoints,
-    ) -> PyRefMut<'py, Self> {
-        contact_points.add_known_nodes(&mut slf.config);
+    ) -> PyRef<'py, Self> {
+        {
+            let mut inner = slf.inner.lock_py_attached(py).unwrap();
+
+            contact_points.clone().add_known_nodes(&mut inner.config);
+            inner.contact_points.extend(contact_points.inner);
```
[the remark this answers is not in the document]
Lorak-mmk: Let's not hide panics.

src/enums.rs
```diff
@@ -220,12 +221,106 @@ impl PyWriteCoalescingDelay {
...
+        client_id: Option<String>,
+    ) -> Self {
+        let mut inner = SelfIdentity::new();
+
+        if let Some(v) = custom_driver_name {
+            inner.set_custom_driver_name(v);
+        } else {
+            inner.set_custom_driver_name("Python-RS Driver");
+        }
+
```
Lorak-mmk, line 264: `ScyllaDB Python-RS Driver` or `Scylla Python-RS Driver` (cc @wprzytula )

Lorak-mmk (changes requested):

src/enums.rs
wprzytula: ScyllaDB Python RS Driver. No hyphens in the branding.

PR #71, opened 2026-04-30
Session builder builtins

### LINKED ISSUE #69: Implement builtins methods for SessionBuilder API

patrycja-ziemkiewicz:
# Changes 
## Protocol-based approach
HostFilter, AddressTranslator, and TimestampGenerator are now defined as Protocols instead of requiring users to inherit from a concrete base class.

This makes custom implementations simpler and more flexible: users only need to provide objects with the expected methods, without explicitly subclassing driver-provided types.

## Authentication remains class-based
Authenticator and AuthenticatorProvider are intentionally kept as classes.
Authentication flows are more complex than host filtering, address translation, or timestamp generation.  Keeping a class-based API should make these implementations easier to structure and extend.
Built-in authentication providers are intentionally left for a separate PR.
## ContactPoint renamed to NodeAddr
ContactPoint has been renamed to NodeAddr to better reflect that this type represents a node address rather than only an initial contact point.
## New built-in implementations
This PR adds the following built-ins:
### Host filters
- AcceptAllHostFilter
- DcHostFilter
 
It also adds support for list-based host filtering, so users can simply provide a list of allowed hosts without manually implementing a custom host filter.
### Timestamp generators
- SimpleTimestampGenerator
- MonotonicTimestampGenerator
### Address translation
This PR adds built-in support for dictionary-based address translation. Users can provide a mapping from original node addresses to translated addresses directly, without implementing a custom translator class.

Fixes: https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/69

## Pre-review checklist

### DIFF
--- a/python/scylla/_rust/policies.pyi
+++ b/python/scylla/_rust/policies.pyi
@@ -133,15 +133,11 @@ class Peer:
     def rack(self) -> Optional[str]: ...
     def __repr__(self) -> str: ...
 
-class HostFilter:
+@runtime_checkable
+class HostFilter(Protocol):
     """
-    Base class for implementing custom host filtering.
-
-    Subclass this and override :meth:`accept` to decide whether a given
-    node should be considered by the driver.
+    Protocol for implementing custom host filtering.
     """
-
-    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def accept(self, peer: Peer) -> bool:
         """
         Decide whether the given peer should be accepted.
@@ -161,3 +157,17 @@ class HostFilter:
         accepting the host.
         """
         ...
+
+class AcceptAllHostFilter:
+    """
+    A host filter that accepts every node in the cluster.
+    """
+    def __init__(self) -> None: ...
+    def accept(self, peer: Peer) -> bool: ...
+
+class DcHostFilter:
+    """
+    A host filter that accepts nodes only from the specified datacenter.
+    """
+    def __init__(self, local_dc: str) -> None: ...
+    def accept(self, peer: Peer) -> bool: ...
--- a/python/scylla/_rust/policies.pyi
+++ b/python/scylla/_rust/policies.pyi
@@ -1,6 +1,7 @@
 import uuid
 from typing import Optional, Any, Protocol, runtime_checkable
 from ipaddress import IPv4Address, IPv6Address
+from datetime import timedelta
 
 class Authenticator:
     """
@@ -63,25 +64,57 @@ class AddressTranslator(Protocol):
         """
         ...
 
-class TimestampGenerator:
+@runtime_checkable
+class TimestampGenerator(Protocol):
     """
-    Base class for implementing custom client-side timestamp generation.
-
-    Subclass this and override :meth:`next_timestamp` to provide custom logic for generating timestamps for requests.
+    Protocol for custom client-side timestamp generation.
     """
-
-    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def next_timestamp(self) -> int:
         """
         Generate the next timestamp for a request.
 
         This method should return an integer representing the timestamp.
 
-        If this method is not overridden or raises an exception, the
+        If this method is not implemented or raises an exception, the
         driver will log the error and fallback to the current system timestamp.
 
         """
         ...
+
+class SimpleTimestampGenerator:
+    """
+    A simple client-side timestamp generator based on the system clock.
+
+    This generator returns the current system time in microseconds since
+    the Unix Epoch (1970-01-01)
+    """
+    def __init__(self) -> None: ...
+    def next_timestamp(self) -> int: ...
+
+class MonotonicTimestampGenerator:
+    """
+    Timestamp generator that guarantees monotonically increasing timestamps.
+
+    Parameters
+    ----------
+    warn_on_drift : bool, default True
+        Whether to log warnings when generated timestamps drift too far from
+        the system clock.
+
+    warning_threshold : float | timedelta, default 1
+        Drift threshold in seconds after which warnings may be emitted.
+
+    warning_interval : float | timedelta, default 1
+        Minimum interval in seconds between drift warnings.
+    """
+
+    def __init__(
+        self,
+        warn_on_drift: bool = True,
+        warning_threshold: float | timedelta = 1,
+        warning_interval: float | timedelta = 1,
+    ) -> None: ...
+    def next_timestamp(self) -> int: ...
 
 class Peer:
     """
--- a/python/scylla/_rust/policies.pyi
+++ b/python/scylla/_rust/policies.pyi
@@ -1,6 +1,5 @@
-import ipaddress
 import uuid
-from typing import Optional, Tuple, Any
+from typing import Optional, Any, Protocol, runtime_checkable
 from ipaddress import IPv4Address, IPv6Address
 
 class Authenticator:
@@ -42,23 +41,25 @@ class UntranslatedPeer:
     @property
     def host_id(self) -> uuid.UUID: ...
     @property
-    def untranslated_address(self) -> Tuple[ipaddress.IPv4Address | ipaddress.IPv6Address, int]: ...
+    def untranslated_address(self) -> tuple[IPv4Address | IPv6Address, int]: ...
     @property
     def datacenter(self) -> Optional[str]: ...
     @property
     def rack(self) -> Optional[str]: ...
     def __repr__(self) -> str: ...
 
-class AddressTranslator:
+@runtime_checkable
+class AddressTranslator(Protocol):
     """
-    Base class for implementing custom address translation.
-    Subclass this to provide your own translation logic.
+    Protocol for custom address translation.
     """
-    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
-    def translate(self, info: UntranslatedPeer) -> Tuple[ipaddress.IPv4Address | ipaddress.IPv6Address, int]:
+    def translate(self, info: UntranslatedPeer) -> str | tuple[str | IPv4Address | IPv6Address, int]:
         """
         Translates a node's address.
-        Must return a tuple of (ip_address, port_integer).
+        Must return a tuple of (ip_address | str, port_integer) or string with valid address and port.
+
+        When returning a string, it should therefore be a numeric IP address
+        plus port (for example ``"127.0.0.1:9042"`` or ``"[::1]:9042"``).
         """
         ...
 
--- a/python/scylla/_rust/session_builder.pyi
+++ b/python/scylla/_rust/session_builder.pyi
@@ -6,8 +6,12 @@ from ipaddress import IPv4Address, IPv6Address
 from .policies import AuthenticatorProvider, AddressTranslator, TimestampGenerator, HostFilter
 from .execution_profile import ExecutionProfile
 from .session import Session
+from datetime import timedelta
+from typing import Any
 
-ContactPoint = str | tuple[str | IPv4Address | IPv6Address, int]
+Address = str | tuple[str | IPv4Address | IPv6Address, int]
+
+TranslationMap = dict[Any, Any]
 
 class SessionBuilder:
     """
@@ -32,7 +36,7 @@ class SessionBuilder:
         """
         ...
 
-    def contact_points(self, contact_points: ContactPoint | Sequence[ContactPoint]) -> SessionBuilder:
+    def contact_points(self, contact_points: Address | Sequence[Address]) -> SessionBuilder:
         """
         Set the contact points used to bootstrap the connection.
 
@@ -111,14 +115,22 @@ class SessionBuilder:
         """
         ...
 
-    def address_translator(self, translator: AddressTranslator) -> SessionBuilder:
+    def address_translator(self, translator: AddressTranslator | TranslationMap) -> SessionBuilder:
         """
-        Registers a custom Python-defined address translator.
+        Registers an address translator for the session.
+
+        The translator can be either a custom Python object implementing the
+        :class:`AddressTranslator` protocol or a static translation mapping.
 
         Parameters
         ----------
-        translator : AddressTranslator
-            An instance of a class inheriting from :class:`AddressTranslator`.
+        translator : AddressTranslator | dict
+            The translation logic to apply. Can be:
+
+            Addresses in the dictionary can be provided as:
+
+            * A string: ``"127.0.0.1:9042"``
+            * A tuple: ``("127.0.0.1", 9042)``, ``(IPv4Address("127.0.0.1"), 9042)``, etc.
 
         Returns
         -------
--- a/python/scylla/_rust/session_builder.pyi
+++ b/python/scylla/_rust/session_builder.pyi
@@ -6,7 +6,6 @@ from ipaddress import IPv4Address, IPv6Address
 from .policies import AuthenticatorProvider, AddressTranslator, TimestampGenerator, HostFilter
 from .execution_profile import ExecutionProfile
 from .session import Session
-from datetime import timedelta
 from typing import Any
 
 Address = str | tuple[str | IPv4Address | IPv6Address, int]
@@ -149,7 +148,8 @@ class SessionBuilder:
         Parameters
         ----------
         generator : TimestampGenerator
-            An instance of a class inheriting from :class:`TimestampGenerator`.
+            A custom Python object implementing the :class:`TimestampGenerator`
+            protocol.
 
         Returns
         -------
--- a/python/scylla/_rust/session_builder.pyi
+++ b/python/scylla/_rust/session_builder.pyi
@@ -157,17 +157,23 @@ class SessionBuilder:
         """
         ...
 
-    def host_filter(self, host_filter: HostFilter) -> SessionBuilder:
+    def host_filter(self, host_filter: HostFilter | Sequence[Address]) -> SessionBuilder:
         """
-        Registers a custom Python-defined host filter.
+        Registers a host filter or a list of allowed addresses.
 
-        The filter is consulted to decide whether a discovered node should be
-        accepted by the driver.
+        This decides whether a discovered node should be accepted by the driver.
+        You can provide a custom filter object for complex logic, or a simple
+        sequence of addresses to act as an allow-list.
 
         Parameters
         ----------
-        host_filter : HostFilter
-            An instance of a class inheriting from :class:`HostFilter`.
+        host_filter : HostFilter | Sequence[Address]
+            If a object implements :class:`HostFilter` protocol, the driver calls
+            its ``accept`` method for each node.
+            If a sequence of addresses, only nodes matching those addresses
+            will be accepted.
+
+            Address = str | tuple[str | IPv4Address | IPv6Address, int]
 
         Returns
         -------
--- a/python/tests/test_session_builder.py
+++ b/python/tests/test_session_builder.py
@@ -56,9 +56,8 @@ async def test_contact_points_invalid_types(item: Any):
     with pytest.raises(SessionConfigError) as excinfo:
         builder.contact_points(item)  # type: ignore[arg-type]
 
-    assert (
-        "Invalid contact points type: expected str | tuple(str, int) | tuple(ipaddress, int) or a sequence of these"
-        in str(excinfo.value.__cause__)
+    assert "Invalid address type: expected str | tuple(str, int) | tuple(ipaddress, int) or a sequence of these" in str(
+        excinfo.value.__cause__
     )
 
 
--- a/python/tests/test_session_builder.py
+++ b/python/tests/test_session_builder.py
@@ -153,7 +153,7 @@ async def test_builtin_user_credentials(ccm_contact_points: list[tuple[str, int]
 Address = ipaddress.IPv4Address | ipaddress.IPv6Address
 
 
-class MockAddressTranslator(AddressTranslator):
+class MockAddressTranslator:
     default_ip: Address
     default_port: int
     call_log: list[UntranslatedPeer]
@@ -169,7 +169,7 @@ class MockAddressTranslator(AddressTranslator):
         return self.default_ip, self.default_port
 
 
-class FailingTranslator(AddressTranslator):
+class FailingTranslator:
     def translate(self, info: UntranslatedPeer) -> tuple[Address, int]:
         raise RuntimeError("Translation Exploded!")
 
@@ -178,6 +178,7 @@ class FailingTranslator(AddressTranslator):
 @pytest.mark.requires_db
 async def test_custom_address_translator_discovery():
     translator = MockAddressTranslator(ipaddress.IPv4Address("127.0.0.2"), 9042)
+    assert isinstance(translator, AddressTranslator)
 
     builder = (
         SessionBuilder()
@@ -202,6 +203,7 @@ async def test_custom_address_translator_discovery():
 @pytest.mark.xfail(reason="Currently, Python exceptions in the translator do not propagate to the driver")
 async def test_address_translator_failing_python_side():
     translator = FailingTranslator()
+    assert isinstance(translator, AddressTranslator)
 
     builder = (
         SessionBuilder()
@@ -214,6 +216,45 @@ async def test_address_translator_failing_python_side():
         await builder.connect()
 
     assert "Translation Exploded" in str(excinfo.value)
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_address_translator_dict_discovery():
+    translator = {
+        (ipaddress.IPv4Address("127.0.0.2"), 9042): (ipaddress.IPv4Address("127.0.0.2"), 9042),
+        ("127.0.0.3", 9042): ("127.0.0.2", 9042),
+        ("127.0.0.4", 9042): ("127.0.0.2", 9042),
+    }
+
+    builder = (
+        SessionBuilder()
+        .contact_points([("127.0.0.2", 9042)])
+        .address_translator(translator)
+        .user("cassandra", "cassandra")
+    )
+
+    session = await builder.connect()
+
+    assert session is not None
+
+
+@pytest.mark.asyncio
+@pytest.mark.requires_db
+async def test_address_translator_dict_invalid():
+    translator = {
+        ("127.0.0.3.3", 9042): ("127.0.0.2.5", 9042),
+    }
+
+    with pytest.raises(SessionConfigError) as excinfo:
+        _ = (
+            SessionBuilder()
+            .contact_points([("127.0.0.2", 9042)])
+            .address_translator(translator)
+            .user("cassandra", "cassandra")
+        )
+
+    assert "invalid socket address syntax" in str(excinfo.value.__cause__)
 
 
 class MockTimestampGenerator(TimestampGenerator):
--- a/python/tests/test_session_builder.py
+++ b/python/tests/test_session_builder.py
@@ -11,9 +11,11 @@ from scylla.policies import (
     AuthenticatorProvider,
     AddressTranslator,
     UntranslatedPeer,
-    TimestampGenerator,
     HostFilter,
     Peer,
+    TimestampGenerator,
+    MonotonicTimestampGenerator,
+    SimpleTimestampGenerator,
 )
 
 from tests.helpers.ccm import (  # pyright: ignore[reportMissingTypeStubs]
@@ -22,6 +24,8 @@ from tests.helpers.ccm import (  # pyright: ignore[reportMissingTypeStubs]
     start_cluster,
     stop_and_remove_cluster,
 )
+
+import time
 
 
 @pytest.mark.asyncio
@@ -257,7 +261,7 @@ async def test_address_translator_dict_invalid():
     assert "invalid socket address syntax" in str(excinfo.value.__cause__)
 
 
-class MockTimestampGenerator(TimestampGenerator):
+class MockTimestampGenerator:
     fixed_ts: int
     called: bool
 
@@ -271,7 +275,7 @@ class MockTimestampGenerator(TimestampGenerator):
         return self.fixed_ts
 
 
-class FailingTimestampGenerator(TimestampGenerator):
+class FailingTimestampGenerator:
     def next_timestamp(self) -> int:
         raise RuntimeError("Timestamp Generation Exploded!")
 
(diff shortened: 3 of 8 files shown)


python/tests/test_session_builder.py
```diff
@@ -347,3 +347,18 @@ async def test_custom_host_filter_fallback_on_failure(
...
+        "127.0.0.1",
+        "::1",
+        ipaddress.IPv4Address("127.0.0.2"),
+        ipaddress.IPv6Address("::2"),
+        None,
+    ],
+)
+def test_local_ip_address_valid_formats(ip: Any):
+    builder = SessionBuilder().local_ip_address(ip)
+    assert isinstance(builder, SessionBuilder)
```
wprzytula: 💭 Can we also verify what happens when we pass invalid data (invalid type or invalid value) to new functions?

python/scylla/_rust/session_builder.pyi
```diff
@@ -6,8 +6,12 @@ from ipaddress import IPv4Address, IPv6Address
 from .policies import AuthenticatorProvider, AddressTranslator, TimestampGenerator, HostFilter
 from .execution_profile import ExecutionProfile
 from .session import Session
+from datetime import timedelta
+from typing import Any
 
-ContactPoint = str | tuple[str | IPv4Address | IPv6Address, int]
+Address = str | tuple[str | IPv4Address | IPv6Address, int]
+
+TranslationMap = dict[Any, Any]
```
wprzytula: 🔧 I dislike that this isn't typed.

python/scylla/_rust/session_builder.pyi
```diff
@@ -111,14 +115,22 @@ class SessionBuilder:
...
+        :class:`AddressTranslator` protocol or a static translation mapping.
 
         Parameters
         ----------
-        translator : AddressTranslator
-            An instance of a class inheriting from :class:`AddressTranslator`.
+        translator : AddressTranslator | dict
+            The translation logic to apply. Can be:
+
+            Addresses in the dictionary can be provided as:
```
wprzytula: > Can be:

🔧 Seems cut off.

python/scylla/_rust/session_builder.pyi
```diff
@@ -111,14 +115,22 @@ class SessionBuilder:
...
+
+            * A string: ``"127.0.0.1:9042"``
+            * A tuple: ``("127.0.0.1", 9042)``, ``(IPv4Address("127.0.0.1"), 9042)``, etc.
```
wprzytula: > `, etc.` 

I don't really have an idea what else can come here. Document it exhaustively, and please provide strict type hints. Otherwise, users have no idea what you support and what not, which is a huge pain in dynamically typed languages.

src/errors.rs
```diff
@@ -469,28 +468,36 @@ pub enum DriverSessionConfigError {
...
+
+    InvalidAddressTranslator {
+        type_name: String,
+    },
+
+    InvalidContactPointAddress {
+        addr: String,
+        source: AddrParseError,
+    },
 }
```
wprzytula, line 601: 🤔 You remove existing docstrings rather than add missing ones? Why??

src/policies.rs
```diff
@@ -156,17 +146,71 @@ impl AddressTranslator for InternalAddressTranslator {
...
-        let result = Python::attach(|py| -> PyResult<(IpAddr, u16)> {
-            let py_trans = self.python_translator.bind(py);
-            let py_peer_info = PyUntranslatedPeer::from(untranslated_peer);
+        Python::attach(|py| -> PyResult<SocketAddr> {
+            let py_trans = self.inner.bind(py);
+            let peer_info = PyUntranslatedPeer::from(untranslated_peer);
+
+            let translated = py_trans
+                .call_method1(intern!(py, "translate"), (peer_info,))?
+                .extract::<ContactPoint>()?;
```
wprzytula: 🚀 Wow, this `intern!` looks super cool! Are we using it everywhere else?
Can it simplify our code in other places, e.g. by removing manual caching using `OnceLock` etc.?

src/policies.rs
```diff
@@ -129,25 +136,8 @@ impl AuthenticatorSession for InternalAuthenticator {
...
-
-    fn translate(&self, _addr: Py<PyUntranslatedPeer>) -> PyResult<(IpAddr, u16)> {
-        Err(PyNotImplementedError::new_err("Method is not implemented"))
-    }
-}
-
-pub(crate) struct InternalAddressTranslator {
-    pub(crate) python_translator: Py<PyAddressTranslator>,
+struct InternalAddressTranslator {
+    inner: Py<PyAny>,
```
wprzytula: 🔧 Missing docstring, the struct looks cryptic.

src/policies.rs
```diff
@@ -156,17 +146,71 @@ impl AddressTranslator for InternalAddressTranslator {
...
+            SocketAddr::try_from(translated).map_err(|e| e.into())
         })
-        .map_err(CustomTranslationError::new)?;
+        .map_err(|e| CustomTranslationError::new(e).into())
+    }
+}
+
+pub(crate) struct AddressTranslatorInput {
+    inner: Arc<dyn AddressTranslator>,
+}
```
wprzytula: 🔧 Missing docstring, the struct looks cryptic.

src/policies.rs
```diff
@@ -156,17 +146,71 @@ impl AddressTranslator for InternalAddressTranslator {
...
+            return Err(DriverSessionConfigError::invalid_address_translator(obj));
+        }
+
+        Ok(Self {
+            inner: Arc::new(InternalAddressTranslator {
+                inner: obj.unbind(),
+            }),
+        })
     }
 }
```
wprzytula: 🔧 Missing comments.

wprzytula: "Address translator can be either a TranslationMap dict or an object implementing AddressTranslator Protocol."
"Here we check for the TranslationMap case"
etc.

src/policies.rs
```diff
@@ -156,17 +146,71 @@ impl AddressTranslator for InternalAddressTranslator {
...
+                .collect::<Result<HashMap<SocketAddr, SocketAddr>, Self::Error>>()?;
+
+            return Ok(Self {
+                inner: Arc::new(map),
+            });
+        }
+
+        if !obj.hasattr(intern!(obj.py(), "translate")).unwrap_or(false) {
+            return Err(DriverSessionConfigError::invalid_address_translator(obj));
+        }
```
wprzytula, line 235: 💭 Here I'm not sure if we want to use `intern!`. This is when we build the `AddressTranslator`, which should only happen once. It's not worth statically allocating `"translate"` string just for oneshot use.

src/policies.rs
wprzytula: The name is weird. Why `Input`?

python/scylla/_rust/policies.pyi
```diff
@@ -1,6 +1,7 @@
 import uuid
```
wprzytula, line 1: commit `Refractor TimeStampGenerator and add builtins` (and all other commit names):
⛏️ It's _refactor_ (like _factor_), not _refractor_ (like _fraction_). Refraction is an optics matter, refactorization is what we're doing 😉

python/scylla/_rust/policies.pyi
```diff
@@ -42,46 +42,80 @@ class UntranslatedPeer:
...
+    warning_interval : int, default 1
+        Minimum interval in seconds between drift warnings.
+    """
+
+    def __init__(
+        self,
+        warn_on_drift: bool = True,
+        warning_threshold: float | timedelta = 1,
+        warning_interval: float | timedelta = 1,
+    ) -> None: ...
```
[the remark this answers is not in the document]
wprzytula: @patrycja-ziemkiewicz This seems legit issue.

python/scylla/_rust/policies.pyi
```diff
@@ -42,23 +41,25 @@ class UntranslatedPeer:
...
+    def translate(self, info: UntranslatedPeer) -> str | tuple[str | IPv4Address | IPv6Address, int]:
         """
         Translates a node's address.
-        Must return a tuple of (ip_address, port_integer).
+        Must return a tuple of (ip_address | str, port_integer) or string with valid address and port.
+
+        When returning a string, it should therefore be a numeric IP address
+        plus port (for example ``"127.0.0.1:9042"`` or ``"[::1]:9042"``).
         """
         ...
```
Lorak-mmk: Why did you change from the normal class to protocol? Class makes more sense to me.

python/scylla/_rust/session_builder.pyi
```diff
@@ -111,14 +115,22 @@ class SessionBuilder:
         """
         ...
 
-    def address_translator(self, translator: AddressTranslator) -> SessionBuilder:
+    def address_translator(self, translator: AddressTranslator | TranslationMap) -> SessionBuilder:
```
Lorak-mmk: I don't see much point in having a special case for the map. You can have a DictAddressTranslator that is constructed from dict, and pass it here.

python/scylla/_rust/policies.pyi
```diff
@@ -63,26 +64,58 @@ class AddressTranslator(Protocol):
...
-    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def next_timestamp(self) -> int:
         """
         Generate the next timestamp for a request.
 
         This method should return an integer representing the timestamp.
 
-        If this method is not overridden or raises an exception, the
+        If this method is not implemented or raises an exception, the
         driver will log the error and fallback to the current system timestamp.
```
Lorak-mmk, line 103: ditto, can we go back to normal classes?

python/scylla/_rust/policies.pyi
```diff
@@ -133,15 +133,11 @@ class Peer:
...
-class HostFilter:
+@runtime_checkable
+class HostFilter(Protocol):
     """
-    Base class for implementing custom host filtering.
-
-    Subclass this and override :meth:`accept` to decide whether a given
-    node should be considered by the driver.
+    Protocol for implementing custom host filtering.
     """
```
Lorak-mmk, line 162: same here.

python/scylla/_rust/session_builder.pyi
```diff
@@ -157,17 +157,23 @@ class SessionBuilder:
...
+    def host_filter(self, host_filter: HostFilter | Sequence[Address]) -> SessionBuilder:
         """
-        Registers a custom Python-defined host filter.
+        Registers a host filter or a list of allowed addresses.
 
-        The filter is consulted to decide whether a discovered node should be
-        accepted by the driver.
+        This decides whether a discovered node should be accepted by the driver.
+        You can provide a custom filter object for complex logic, or a simple
+        sequence of addresses to act as an allow-list.
```
Lorak-mmk: Same here as for address translator: we can have AllowlistHostFilter that takes a sequence of addresses. No need for special case in public API

Lorak-mmk (changes requested):

python/scylla/_rust/policies.pyi
patrycja-ziemkiewicz: I believe for this use case protocols are more pythonic. What do you think @jakub-graczyk?

jakub-graczyk: I'm in favor of protocol personally. There isn't any inheritance or complexity here to warrant a class.

python/scylla/_rust/session_builder.pyi
patrycja-ziemkiewicz: I do not know how to make it typed. This is the same problem @paulinaczajkowska encountered in serialization docstrings.

src/policies.rs
patrycja-ziemkiewicz: No, we will have to adjust the code to use it in all applicable places. Created issue for that https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/115. Also it can simplify the logic only for connected to storing python strings.

src/policies.rs
patrycja-ziemkiewicz: Input might be not the best naming, I will change it to PyAddressTranslator as it is a struct representing python objects implementing AdressTranslatorTrait.

src/policies.rs
```diff
@@ -129,57 +133,134 @@ impl AuthenticatorSession for InternalAuthenticator {
...
+
+        Ok(Self {
+            inner: Arc::new(CustomAddressTranslator {
+                inner: obj.unbind(),
+            }),
+        })
+    }
+}
+/// Python representation of an untranslated peer address, exposing host_id, untranslated_address,
```
wprzytula, line 294: 🔧 Missing newline.

python/scylla/_rust/policies.pyi
```diff
@@ -2,6 +2,8 @@ import uuid
 from ipaddress import IPv4Address, IPv6Address
 from typing import Any, Optional, Protocol, runtime_checkable
 
+Address = str | tuple[str | IPv4Address | IPv6Address, int]
```
wprzytula: 🔧 Deduplicate with `session_builder.pyi`.

python/scylla/_rust/policies.pyi
```diff
@@ -63,7 +65,28 @@ class AddressTranslator(Protocol):
...
+
+    * A string: ``"127.0.0.1:9042"``
+    * A tuple: ``("127.0.0.1", 9042)``, ``(IPv4Address("127.0.0.1"), 9042)``, ``(IPv6Address("::1"), 9042)``.
+    """
+
+    def __init__(self, translation_map: dict[Any, Any]) -> None: ...
+    def translate(self, info: UntranslatedPeer) -> Address: ...
+
+@runtime_checkable
+class TimestampGenerator(Protocol):
```
wprzytula, line 92: ♻️ This shouldn't belong to this commit.

src/session_builder.rs
```diff
@@ -695,6 +695,19 @@ impl<'py> FromPyObject<'_, 'py> for ContactPoint {
...
+
+    fn try_from(value: ContactPoint) -> Result<Self, Self::Error> {
+        match value {
+            ContactPoint::SocketAddr(addr) => Ok(addr),
+            ContactPoint::Host(addr_str) => SocketAddr::from_str(&addr_str).map_err(|reason| {
+                DriverSessionConfigError::invalid_contact_point_addr(addr_str, reason)
+            }),
+        }
+    }
+}
```
wprzytula: 💭 Does this handle DNS hostnames?

patrycja-ziemkiewicz: No it does not. I believe We can leave it like this and mention it in docstrings.

src/errors.rs
```diff
@@ -507,6 +508,12 @@ pub enum DriverSessionConfigError {
     InvalidAddressTranslator {
         type_name: String,
     },
+
+    /// A NodeAddr::Host string could not be parsed into a SocketAddr.
+    InvalidContactPointAddress {
+        addr: String,
+        source: AddrParseError,
+    },
```
Lorak-mmk, line 600: Commit: "Add DictAddressTranslator"

The structure of the errors added in this commit is weird, and I strongly suspect also wrong.
`DriverSessionConfigError` are errors that are supposed to be returned when creating session, right? You added a variant that can never happen there, only for the purposes of converting it to `DriverAddressTranslationError`.


In `DriverAddressTranslationError::invalid_node_address_item` you are also accepting anonymous `PyErr` instead of some strongly typed object, so this weirdness is harder to spot.

I see two approaches:
- In `PyDictAddressTranslator::new` don't use ContactPoint, and perform conversion to SocketAddre yourself. Imo this makes a lot of sense - ContactPoint is not something that is connected to address translator, so its usage by itself is confusing.
- Remove `impl TryFrom<ContactPoint> for SocketAddr`, perform the match on ContactPoint in `PyDictAddressTranslator::new`, which also allows you to drop the new error variant in `DriverSessionConfigError`

python/scylla/_rust/policies.pyi
```diff
@@ -91,21 +92,18 @@ class DictAddressTranslator:
...
-class TimestampGenerator:
+@runtime_checkable
+class TimestampGenerator(Protocol):
     """
-    Base class for implementing custom client-side timestamp generation.
-
-    Subclass this and override :meth:`next_timestamp` to provide custom logic for generating timestamps for requests.
+    Protocol for custom client-side timestamp generation.
     """
-
```
Lorak-mmk, line 65: TimestampGenerator is a hot-peth entity (used when executing requests). We should understand if base-class vs protocol approaches have some performance differences, and if so choose the faster one.

src/policies.rs
```diff
@@ -415,6 +401,36 @@ impl TimestampGenerator for InternalTimestampGenerator {
     }
 }
 
+/// Python-facing input type for timestamp generator. Extracts from built-in `PyMonotonicTimestampGenerator`,
+/// `PySimpleTimestampGenerator`, or wraps any Python object with a `next_timestamp` method
+/// as a `CustomTimestampGenerator`.
+pub(crate) struct PyTimestampGenerator {
+    inner: Arc<dyn TimestampGenerator>,
+}
```
Lorak-mmk, line 442: How is it python facing if it has no `pyclass`?

src/policies.rs
```diff
@@ -129,48 +133,65 @@ impl AuthenticatorSession for InternalAuthenticator {
...
 }
 
-#[pyclass(get_all, name = "UntranslatedPeer", frozen)]
+impl<'py> FromPyObject<'_, 'py> for PyAddressTranslator {
+    type Error = DriverSessionConfigError;
+
+    fn extract(obj: Borrowed<'_, 'py, PyAny>) -> Result<Self, Self::Error> {
+        if !obj.hasattr(intern!(obj.py(), "translate")).unwrap_or(false) {
+            return Err(DriverSessionConfigError::invalid_address_translator(obj));
+        }
```
Lorak-mmk, line 235: Is there no better way to check if something satisfies a protocol? This attribute could be a non-function.

Pyo3 has `Mapping` object (which is a protocol on Python side afaik). How are they doing this?

src/policies.rs
```diff
@@ -417,6 +420,11 @@ impl<'py> FromPyObject<'_, 'py> for PyTimestampGenerator {
     type Error = DriverSessionConfigError;
 
     fn extract(obj: Borrowed<'_, 'py, PyAny>) -> Result<Self, Self::Error> {
+        if let Ok(monotonic) = obj.cast::<PyMonotonicTimestampGenerator>() {
+            return Ok(Self {
+                inner: monotonic.get().inner.clone(),
+            });
+        }
```
Lorak-mmk, line 464: If you are cloning an Arc or other smart pointer, please use `Arc::clone` (or a respective method for other smart pointer) rather than `.clone()`. It makes it more explicit that it is a cheap clone.

src/policies.rs
```diff
@@ -426,6 +426,12 @@ impl<'py> FromPyObject<'_, 'py> for PyTimestampGenerator {
             });
         }
 
+        if let Ok(simple) = obj.cast::<PySimpleTimestampGenerator>() {
+            return Ok(Self {
+                inner: simple.get().inner.clone(),
+            });
+        }
+
```
Lorak-mmk, line 465: ditto

src/errors.rs
Lorak-mmk, line 1: Commit: "Refactor ContactPoint struct"

Why? "Addr" is already an overloaded term (we have IpAddr, SocketAddr). ContactPoint is also no always an address - it can be a DNS name.
Old Python driver uses `contact_points` as the name of the argument, so it would be more familiar to users.

Please also note that contact point does not have to correspond to a single node - a DNS name can map to multiple IPs, of different nodes, so calling it NodeAddr is misleading.

src/policies.rs
patrycja-ziemkiewicz: This means that it stores input provided by the python user, not that this class itself is an python object.

src/policies.rs
patrycja-ziemkiewicz: There isn't a better way to do this at runtime. Python's isinstance() against a @runtime_checkable Protocol just does hasattr checks for each required method/attribute, it doesn't verify callability, signatures, or return types. That validation only happens in static type checkers.
PyO3's PyMapping is a different case, it uses CPython's C API (PyMapping_Check), which checks whether the type's C struct has the tp_as_mapping slot filled in. That only works for a fixed set of built-in protocols that have dedicated slots in CPython's type struct. User-defined protocols have no such mechanism.

src/errors.rs
patrycja-ziemkiewicz: I thought NodeAddr would be a better name since this struct is now also used in the address translator and host filter, but you're right that it's misleading. I'll switch it back to ContactPoint.

python/scylla/_rust/policies.pyi
patrycja-ziemkiewicz: There will be no performance differences. However, if a user implements a custom TimestampGenerator policy, it will be much slower than using an exposed one. This is because we would need to attach to the GIL each time and call the Python function. That being said, the overhead is identical for both the protocol and the base class.

src/session_builder.rs
```diff
@@ -645,67 +643,18 @@ impl<'py> FromPyObject<'_, 'py> for PyDuration {
...
-
-        Err(DriverSessionConfigError::contact_point_type_error(obj))
-    }
-}
-
 #[derive(Clone, Default)]
 struct ContactPoints {
-    inner: Vec<ContactPoint>,
+    inner: Vec<ParsedAddress>,
 }
```
Lorak-mmk, line 639: Commit: "Refactor Parsing Address"

Why "ContactPoints" still exists? You added "ParsedAddressList' which also stores a Vec of ParsedAddress, so why do we need both?

src/session_builder.rs
```diff
@@ -461,7 +459,7 @@ struct PySessionBuilderConfig {
     #[pyo3(get)]
     pub execution_profile: Py<ExecutionProfile>,
     #[pyo3(get)]
-    pub contact_points: Vec<ContactPoint>,
+    pub contact_points: Vec<ParsedAddress>,
```
Lorak-mmk, line 451: ditto

src/policies.rs
```diff
@@ -129,48 +133,65 @@ impl AuthenticatorSession for InternalAuthenticator {
...
-        Err(PyNotImplementedError::new_err("Method is not implemented"))
     }
 }
 
-pub(crate) struct InternalAddressTranslator {
-    pub(crate) python_translator: Py<PyAddressTranslator>,
+/// Stores a Python object with a `translate` method (user's custom implementation)
+/// and implements the Rust `AddressTranslator` trait by delegating to that Python object.
+struct CustomAddressTranslator {
+    inner: Py<PyAny>,
```
Lorak-mmk, line 188: This is, for me, an important downside of protocol-based approach. Before this fields had type `Py<PyAddressTranslator>`, now itsa `Py<PyAny>`. We lose strong typing :cry:

src/policies.rs
```diff
@@ -129,48 +133,65 @@ impl AuthenticatorSession for InternalAuthenticator {
...
+                .extract::<ContactPoint>()?;
 
-            py_trans
-                .call_method1("translate", (py_peer_info,))?
-                .extract::<(IpAddr, u16)>()
+            SocketAddr::try_from(translated).map_err(|e| e.into())
         })
-        .map_err(CustomTranslationError::new)?;
+        .map_err(|e| CustomTranslationError::new(e).into())
+    }
```
Lorak-mmk, line 208: Why before we used a string literal, and now we need `intern!`?

src/errors.rs
```diff
@@ -658,14 +658,64 @@ impl From<DriverSessionConfigError> for PyErr {
...
+#[must_use]
+pub enum DriverAddressTranslationError {
+    TranslationError {
+        source: Box<TranslationError>,
+    },
+
+    InvalidAddress {
+        index: usize,
+        source: AddressParseError,
+    },
```
Lorak-mmk: `DriverAddressTranslationError` sounds likes something that should be general, not specific to any single translator. `index` field here OTOH seems pretty specific to dict address translator.

src/policies.rs
patrycja-ziemkiewicz: We do not need it, but it is a optimization: https://pyo3.rs/main/doc/pyo3/macro.intern

Lorak-mmk: Ah, the name argument is `IntoPyObject<'py, Target = PyString>`, so if we pass a literal then a conversion takes place. It makes sense now, thanks.

src/session_builder.rs
patrycja-ziemkiewicz: I did not want to mix types. Contact point has its internal method add_known_nodes which takes SessionBuilderConfig as argument. This method does not make sense on ParsedAddress struct.

src/errors.rs
patrycja-ziemkiewicz: DriverAddressTranslationError is meant to be general across translators, and we can add other error variants as needed later. I'll rename InvalidAddress to InvalidAddressAtIndex to make this one clearer.

src/policies.rs
Lorak-mmk: Don't change that ofc (keep the protocols), we can re-evaluiate in the future if needed.

src/session_builder.rs
Lorak-mmk: OK

src/session_builder.rs
```diff
@@ -94,17 +93,16 @@ impl SessionBuilder {
...
+    ) -> Result<PyRef<'py, Self>, DriverSessionConfigError> {
         {
             let mut inner = slf.inner.lock_py_attached(py).unwrap();
-            inner.address_translator = Some(translator.clone().into());
-            inner.config.address_translator = Some(Arc::new(InternalAddressTranslator {
-                python_translator: translator,
-            }));
+            let translator = py_translator.extract::<PyAddressTranslator>(py)?;
+            inner.address_translator = Some(py_translator.clone());
+            inner.config.address_translator = Some(translator.into_inner());
```
Lorak-mmk: You are calling `py_translator.extract::<PyAddressTranslator>`, so why can't you still accept `Py<PyAddressTranslator>` as an argument, as it was before your change?

src/session_builder.rs
```diff
@@ -124,17 +124,17 @@ impl SessionBuilder {
...
         {
             let mut inner = slf.inner.lock_py_attached(py).unwrap();
-            inner.host_filter = Some(host_filter.clone().into());
-            inner.config.host_filter = Some(Arc::new(InternalHostFilter {
-                py_host_filter: host_filter,
-            }));
+            let host_filter = py_host_filter.extract::<PyHostFilter>(py)?;
+
+            inner.host_filter = Some(py_host_filter.clone());
+            inner.config.host_filter = Some(host_filter.into_inner());
```
Lorak-mmk: ditto

src/session_builder.rs
```diff
@@ -108,17 +108,17 @@ impl SessionBuilder {
...
         {
             let mut inner = slf.inner.lock_py_attached(py).unwrap();
-            inner.timestamp_generator = Some(generator.clone().into());
-            inner.config.timestamp_generator = Some(Arc::new(InternalTimestampGenerator {
-                py_timestamp_generator: generator,
-            }));
+            let generator = py_generator.extract::<PyTimestampGenerator>(py)?;
+            inner.timestamp_generator = Some(py_generator.clone());
+
+            inner.config.timestamp_generator = Some(generator.into_inner());
```
Lorak-mmk: and here

src/session_builder.rs
patrycja-ziemkiewicz: PyAddressTranslator stores internally: `Arc<dyn AddresssTranslator>`, I want config to store the exact object user have passed to us. To be able to do that I need to get from somewhere unchanged `Py<PyAny>` user have passed. I can do that either by having another field in the `PyAddressTranslator` with value passed by user or the same as I am doing now.

Lorak-mmk: I didn't really understand your comment, so I tried to apply my idea locally.

The way I understand the issue now is that we want to have a getter on `PySessionBuilderConfig`, so the user can read the address translator. We could put `Py<PyAuthenticatorProvider>` as the Rust-side field. The downside is that we would then have to introduce a new allocation - because we can't use `Py<PyAuthenticatorProvider>` as a function argument.

It seems like quite a basic problem in PyO3 usage. Is there really no better solution?

patrycja-ziemkiewicz: I do not understand now. We cannot do `Py<PyAddressTranslator>` as function argument because `PyAddressTranslator` is not a pyclass. It is only struct used by us internally to extract value given by user and try to cast it to some of the builtin policies. If it fails We assume that user have provided a custom policy. The problem is that this struct inside stores `Arc<dyn AddressTranslator>` which can be either our `CustomAddressTranslator` or some internal rust struct implementing this trait (e.g. `HashMap<SocketAddr, SocketAddr>`). If user want to inspect the AddressTranslator he set on `SessionBuilder` We want to return to him the exact same object he have provided.

Lorak-mmk: One idea is to have an util like
```rust
pub(crate) struct WithOriginalPyObject<T> {
    pub(crate) original: Py<PyAny>,
    pub(crate) extracted: T,
}

impl<'a, 'py, T> FromPyObject<'a, 'py> for PyCloned<T>
where
    T: FromPyObject<'a, 'py>,
{
    type Error = <T as FromPyObject<'a, 'py>>::Error;

    fn extract(obj: Borrowed<'a, 'py, PyAny>) -> Result<Self, Self::Error> {
        let original = obj.to_owned().unbind();
        Ok(Self {
            extracted: obj.extract::<T>()?,
            original,
        })
    }
}
```

Then this function becomes
```rust
    fn address_translator<'py>(
        slf: PyRef<'py, Self>,
        py: Python<'py>,
        translator: WithOriginalPyObject<PyAddressTranslator>,
    ) -> Result<PyRef<'py, Self>, DriverSessionConfigError> {
        {
            let mut inner = slf.inner.lock_py_attached(py).unwrap();
            inner.address_translator = Some(translator.original.clone());
            inner.config.address_translator = Some(translator.extracted.into_inner());
        }

        Ok(slf)
    }
```

It still doesn't address the problem of having 
```rust
    #[pyo3(get)]
    pub address_translator: Option<Py<PyAny>>,
```

in the object, but it is still an improvement.

patrycja-ziemkiewicz: Yes, that is exacly what I have proposed here. 

> I can do that either by having another field in the `PyAddressTranslator` with value passed by user 

If you prefer this approach I can implement it

Lorak-mmk: Please do, but also notice that my solution doesn't add a field to PyAddressTranslator, so it can be reused for timestamp generator and host filter.

PR #72, opened 2026-05-04
Implement TLS/SSL context configuration for SessionBuilder

### LINKED ISSUE #70: Introduce custom TLS Implementation for Python Driver
**Description**
We need to introduce TLS support for the driver to securely connect to ScyllaDB clusters. After investigating the FFI boundaries between Python and our Rust core, we have determined that directly integrating with Python's standard `ssl.SSLContext` is not a viable or safe approach for our use case. 

Instead, we need to implement a custom `TlsContext` layer that allows the Rust core to manage its own cryptographic state securely, completely decoupled from Python's internal `ssl` module.

**Why not use Python's `ssl.SSLContext`? (Known Limitations)**
During our investigation into passing Python's SSL objects directly to Rust, we identified two major architectural blockers:

1. **ABI Instability & Memory Ownership (Safety Risks):** CPython's internal `PySSLContext` struct is strictly private. Extracting the raw OpenSSL C-pointer (`SSL_CTX`) from it requires unsafe pointer arithmetic. Because this layout changes across minor Python versions, relying on it guarantees ABI breakages. Furthermore, sharing this pointer causes conflicts with Python's Garbage Collector, leading to use-after-free errors and unpredictable segmentation faults.

2. **mTLS is Impossible via Getters (Feature Limitations):** While we could theoretically map simple configuration flags (like `verify_mode`) from Python's context over to Rust, Python's `ssl` module completely swallows the client's private key upon loading. There is no getter available to extract the private key back out. Without the private key, the Rust core cannot authenticate, making Mutual TLS (mTLS) impossible to support if we rely on Python's native context.

**Proposed Direction**
* Do not attempt to extract pointers from or map Python's `ssl.SSLContext`.
* Create a custom `TlsContext` class exposed to Python.
* Allow the Rust driver to safely build, isolate, and manage its own internal TLS context without interacting with CPython internals.

pasinskik:
## Pre-review checklist

- [X] All commit messages clearly explain what they change and why.
- [X] I added relevant tests for new features and bug fixes.
- [X] All commits compile, pass static checks and pass test.
- [X] PR description sums up the changes and reasons why they should be introduced.
- [X] I have provided docstrings for the public items that I want to introduce.
- [X] I added appropriate `Fixes:` annotations to PR description.

## Summary

This PR introduces native TLS/SSL support to the Python driver. 

It adds Python-facing TLS configuration wrappers and exposes them through:
- `scylla.tls.TlsContextBuilder`
- `SessionBuilder.tls_context()`

## Motivation

Users need the ability to establish secure connections to ScyllaDB clusters. While the underlying Rust driver supports TLS via OpenSSL, the Python-RS driver did not expose a way to configure it.

Exposing this configuration allows for both standard server-side verification and Mutual TLS (mTLS), which are critical requirements for enterprise and production deployments.

## Changes

- Added native Rust `TlsContext` implementation backed by OpenSSL `SslContext`.
- Exposed `TlsContext` to Python with support for:
  - custom CA certificates,
  - client certificate and private key configuration,
  - optional peer verification.
- Added validation for PEM-encoded CA certificates, client certificate chains,
  and private keys.
- Added support for client certificate chains, using the first certificate as
  the leaf certificate and the remaining certificates as intermediates.
- Added private key validation with `check_private_key()`.
- Added parameter validation requiring client certificate and key to be provided
  together for mTLS.
- Configured default system trust paths only when peer verification is enabled
  and no custom CA bundle is provided.
- Added pure-Python `TlsContextBuilder` API for loading TLS material from files
  or in-memory PEM data.
- Added support for mixing file-based and in-memory certificate sources.
- Added type stubs for the public TLS API.
- Added TLS usage example demonstrating:
  - server certificate verification from disk,
  - mutual TLS using in-memory PEM data,
  - mixed CA file and in-memory client credentials.

## Why this design

The goal of this PR is to expose TLS configuration safely without leaking FFI boundary complexity to the user.

For that reason:
- **Stateful Python Builder:** Rather than a massive one-shot constructor, `TlsContextBuilder` directly mirrors Python's native `ssl.SSLContext` design. Users can incrementally configure their context using `load_verify_locations` and `load_cert_chain`. This allows users to seamlessly mix and match CA certificates loaded from disk with mTLS client credentials pulled into memory from a vault. 
- **Explicit FFI Boundary:** The user explicitly compiles their configuration by calling `builder.build()`, which securely hands the raw bytes to the native Rust layer exactly once.
- **Robust Rust Cryptography:** The Rust layer acts purely as a secure consumer of bytes. It uses OpenSSL stacks (`stack_from_pem`) to properly process multi-certificate CA bundles and intermediate client chains, and automatically injects system trust roots if standard server verification is requested.

Fixes: #70


Cargo.toml
```diff
@@ -28,3 +28,4 @@ num-bigint = "0.4"
 async-trait = "0.1.89"
 log = "0.4.29"
 pyo3-log = "0.13.3"
+openssl = "0.10"
```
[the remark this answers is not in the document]
pasinskik: I will change `openssl = "0.10"` to `openssl = { version = "0.10", features = ["vendored"] }`

src/tls.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn new(
+        ca_pem: Option<&[u8]>,
+        cert_pem: Option<&[u8]>,
+        key_pem: Option<&[u8]>,
+        verify_peer: bool,
+    ) -> Result<Self, DriverTlsError> {
+        let mut builder =
+            SslContextBuilder::new(SslMethod::tls()).map_err(DriverTlsError::openssl_error)?;
+
+        if verify_peer {
```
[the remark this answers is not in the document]
pasinskik: I have updated the implementation to call set_default_verify_paths() so system trust roots are loaded automatically.

src/tls.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        if let Some(ca_bytes) = ca_pem {
+            let ca_cert = X509::from_pem(ca_bytes).map_err(|e| {
+                DriverTlsError::invalid_parameters(format!("Invalid CA PEM: {}", e))
+            })?;
+            builder
+                .cert_store_mut()
+                .add_cert(ca_cert)
+                .map_err(DriverTlsError::openssl_error)?;
```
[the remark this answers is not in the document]
pasinskik: The CA loader now iterates through the stack to add all bundle certificates to the store.

src/tls.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        match (cert_pem, key_pem) {
+            (Some(cert_bytes), Some(key_bytes)) => {
+                let cert = X509::from_pem(cert_bytes).map_err(|e| {
+                    DriverTlsError::invalid_parameters(format!("Invalid Cert PEM: {}", e))
+                })?;
+                builder
+                    .set_certificate(&cert)
+                    .map_err(DriverTlsError::openssl_error)?;
```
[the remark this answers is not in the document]
pasinskik: The mTLS loader now cleanly separates the leaf certificate from its intermediates (loading the latter via `add_extra_chain_cert`).

src/tls.rs
```diff
@@ -0,0 +1,106 @@
...
+        if let Some(ca_bytes) = ca_pem {
+            let ca_certs = X509::stack_from_pem(ca_bytes).map_err(|e| {
+                DriverTlsError::invalid_parameters(format!("Invalid CA PEM: {}", e))
+            })?;
+            for cert in ca_certs {
+                builder
+                    .cert_store_mut()
+                    .add_cert(cert)
+                    .map_err(DriverTlsError::openssl_error)?;
+            }
```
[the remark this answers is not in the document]
pasinskik: I've opted for the second suggestion: explicit CA configuration now overrides the system defaults. The builder now only calls set_default_verify_paths() in an else block when ca_pem is completely omitted. This safely avoids duplicate errors and aligns perfectly with standard TLS client behavior.

Cargo.toml
```diff
@@ -28,3 +28,4 @@ num-bigint = "0.4"
 async-trait = "0.1.89"
 log = "0.4.29"
 pyo3-log = "0.13.3"
+openssl = { version = "0.10", features = ["vendored"] }
```
Lorak-mmk: Do we really need vendored? It will compile openssl from source, which will be slow. It WILL impact users. Requiring openssl to be installed in the system doesn't seem problematic to me.

python/scylla/_rust/session_builder.pyi
```diff
@@ -162,3 +163,31 @@ class SessionBuilder:
...
+
+    def tls_context(self, tls_context: TlsContext) -> SessionBuilder:
+        """
+        Configures the session to use the provided TLS/SSL context.
+
+        Parameters
+        ----------
+        tls_context : TlsContext
+            The native Rust TLS context created by the :class:`TlsContext` builder in tls.py.
+
```
Lorak-mmk, line 474: I see no point in mentioning Rust here. This is a Python driver and users shouldn't need to care what is inside.

python/scylla/_rust/session_builder.pyi
```diff
@@ -162,3 +163,31 @@ class SessionBuilder:
...
+        Explicitly disables TLS for the session.
+
+        This is the default behavior, but this method can be used to override
+        a previously set TLS context on this builder instance.
+
+        Returns
+        -------
+        SessionBuilder
+        """
+        ...
```
Lorak-mmk: Why a separate method instead of accepting `None` in tls_context?

python/scylla/_rust/tls.pyi
```diff
@@ -0,0 +1,16 @@
...
+    produces this native context after handling Python-specific logic (like file I/O).
+    """
+    def __init__(
+        self,
+        *,
+        ca_pem: bytes | None = None,
+        cert_pem: bytes | None = None,
+        key_pem: bytes | None = None,
+        verify_peer: bool = True,
+    ) -> None: ...
```
Lorak-mmk, line 26: This doc comment is imo no useful to the user.
What does it even mean that the context is "natively compiled"?
Why user-facing doc mentions Rust Driver here?

Lorak-mmk (changes requested):
Not a full review, just a few comments.
Have you thought about how to also enable rustls usage?

Cargo.toml
wprzytula: One important consideration is whether to link to the system-present OpenSSL library dynamically (at runtime) or statically (at build time). The same choice is open in CPP RS Driver.

python/scylla/_rust/tls.pyi
```diff
@@ -0,0 +1,14 @@
...
+    This is the actual TLS configuration object. `TlsContextBuilder` builds this object, and then it is passed to the session builder to configure TLS for the session.
+    """
+    def __init__(
+        self,
+        *,
+        ca_pem: bytes | None = None,
+        cert_pem: bytes | None = None,
+        key_pem: bytes | None = None,
+        verify_peer: bool = True,
+    ) -> None: ...
```
Lorak-mmk: What is the plan for adding more configuration options? More Optional parameters to constructor?

python/scylla/tls.py
```diff
@@ -0,0 +1,76 @@
...
+    This API allows users to safely mix and match certificates loaded from disk
+    and certificates loaded from memory before building the final context.
+    """
+
+    def __init__(self, verify_peer: bool = True):
+        self._verify_peer = verify_peer
+        self._ca_pem: bytes | None = None
+        self._cert_pem: bytes | None = None
+        self._key_pem: bytes | None = None
+
```
Lorak-mmk, line 36: Could you define the field types on the class, not inside the constructor?

python/scylla/tls.py
```diff
@@ -0,0 +1,76 @@
+from pathlib import Path
+
+from ._rust.tls import TlsContext  # pyright: ignore[reportMissingModuleSource]
+
+__all__ = ["TlsContextBuilder"]
+
```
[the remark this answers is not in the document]
Lorak-mmk: @pasinskik Is that comment not correct? This is how I understand `__all__` as well.

python/tests/test_tls.py
```diff
@@ -0,0 +1,242 @@
...
+                "ca.key",
+                "-out",
+                "ca.crt",
+                "-days",
+                "1",
+                "-nodes",
+                "-subj",
+                "/CN=Test-CA",
+            ]
+        )
```
Lorak-mmk: Is there no other way to do this than calling `openssl` CLI?
In Rust Driver we use `rcgen` to create such temporary certs. I find it hard to believe there is nothing similar for Python.

src/tls.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn new(
+        ca_pem: Option<&[u8]>,
+        cert_pem: Option<&[u8]>,
+        key_pem: Option<&[u8]>,
+        verify_peer: bool,
+    ) -> Result<Self, DriverTlsError> {
+        let mut builder = SslContextBuilder::new(SslMethod::tls_client())
+            .map_err(DriverTlsError::openssl_error)?;
+
+        if verify_peer {
```
Lorak-mmk: Could you try to configure openssl using `openssl::ssl::SslConnector`?
Afaik it has better defaults than SslContextBuilder.

python/tests/test_tls.py
```diff
@@ -0,0 +1,242 @@
...
+            "require_client_auth": True,  # Forces the Python driver to provide a client cert
+            "certificate": str(certs_dir / "server.crt"),
+            "keyfile": str(certs_dir / "server.key"),
+            "truststore": str(certs_dir / "ca.crt"),
+        }
+    }
+
+    cluster = create_scylla_cluster(
+        name="tls_cluster",
+        scylla_version="release:6.2.2",
```
Lorak-mmk: Please explain why did you use such terribly old version.

I'd like to have a response to this comment, not just a code change.

Lorak-mmk: Btw our CCM integration should receive version from the outside, test should not decide on a version to run.

Lorak-mmk (changes requested):

python/scylla/_rust/tls.pyi
pasinskik: I've changed the approach. The plan is to keep the constructor parameterless and add dedicated builder methods for new TLS options, rather than growing it with more optional parameters.

python/scylla/tls.py
pasinskik: Yes, I’ve moved the field annotations to the class body and left only the assignments in `__init__`.

python/tests/test_tls.py
pasinskik: > Please explain why did you use such terribly old version.
> 
> I'd like to have a response to this comment, not just a code change.

I've used this version because there's a test in `test_session_builder.py` that also uses this version.

PR #101, opened 2026-05-27
add: documentation for getting started page

jakub-graczyk:
Getting started page documentation.

### DIFF
--- a/docs/source/topics/getting-started.md
+++ b/docs/source/topics/getting-started.md
@@ -1,3 +1,384 @@
 # Getting Started
 
-TBD.
+First, make sure the driver has been properly installed. See [Installation](installation.md).
+
+> **Note:** This driver is **async-first** and requires Python's `asyncio`. All network operations - connecting, executing statements, fetching results - are coroutines, which must be awaited inside an `async` function.
+
+## Connecting to a Cluster
+
+Before executing any statements you need to establish a connection to a ScyllaDB cluster. This is done using `SessionBuilder`, which provides a chainable API for configuring the connection before calling `connect()`.
+
+### Basic connection
+
+The simplest way to create a `Session` is to provide a single contact point. The driver will automatically discover the rest of the cluster:
+
+```python
+import asyncio
+from scylla.session_builder import SessionBuilder
+
+async def main():
+    builder = SessionBuilder().contact_points("127.0.0.2")
+    session = await builder.connect()
+
+asyncio.run(main())
+```
+
+### Multiple contact points
+
+You can pass a list of initial contact points. The driver needs only one node to discover the rest of the cluster automatically:
+
+```python
+builder = SessionBuilder().contact_points(["127.0.0.2", "127.0.0.3"])
+session = await builder.connect()
+```
+
+Each contact point can be provided as:
+- a `str` - e.g. `"127.0.0.2"` or `"127.0.0.2:9042"`
+- a `tuple[str, int]` - e.g. `("127.0.0.2", 9042)`
+- a `tuple[IPv4Address | IPv6Address, int]`
+
+### Specifying a non-standard port
+
+```python
+builder = SessionBuilder().contact_points([("127.0.0.2", 9142), ("127.0.0.3", 9142)])
+session = await builder.connect()
+```
+
+### Authentication
+
+Use `.user()` for plain-text authentication:
+
+```python
+builder = SessionBuilder().contact_points("127.0.0.2").user("username", "password")
+session = await builder.connect()
+```
+
+## Working with Keyspaces
+
+To set a default keyspace for the session a `use_keyspace` method is available:
+
+```python
+await session.execute(
+    "CREATE KEYSPACE IF NOT EXISTS my_ks "
+    "WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}"
+)
+await session.use_keyspace("my_ks")
+```
+
+All subsequent requests will use `my_ks` as the default keyspace for unqualified table names.
+
+## Executing Statements
+
+Use `session.execute()` to run a CQL statement. It is a coroutine - always `await` it:
+
+```python
+await session.execute(
+    "CREATE TABLE IF NOT EXISTS users (id int PRIMARY KEY, name text, age int)"
+)
+```
+
+### Passing parameters
+
+Use `?` as the placeholder for bound parameters. Supply parameter values as a list or tuple:
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)",
+    (1, "Alice", 30)
+)
+```
+
+### Reading rows
+
+Rows are returned in pages as `dict[str, Any]` by default. Iterate over all rows in all pages with transparent async paging using `async for`:
+
+```python
+result = await session.execute("SELECT id, name, age FROM users")
+async for row in result:
+    print(row["id"], row["name"], row["age"])
+```
+
+### Concurrent execution
+
+Because the driver is async, you can execute many statements concurrently using `asyncio.gather`. This is the recommended approach for high-throughput workloads:
+
+```python
+coroutines = [
+    session.execute("INSERT INTO users (id, name, age) VALUES (?, ?, ?)", [i, f"user_{i}", i % 100])
+    for i in range(1000)
+]
+await asyncio.gather(*coroutines)
+```
+
+> Awaiting each `execute()` call sequentially in a loop is suboptimal - always prefer `asyncio.gather` or any other way of non-blocking execution when inserting or querying large amounts of data.
+
+## Prepared Statements
+
+Prepared statements are parsed by ScyllaDB once and cached. Every subsequent execution only sends the bound parameter values (and prepared statement id, which is a small hash instead of the full statement string to yet be parsed), reducing network traffic and CPU usage.
+
+Prepare a statement with `session.prepare()`:
+
+```python
+insert_statement = await session.prepare(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)"
+)
+```
+
+Execute it as many times as needed:
+
+```python
+await session.execute(insert_statement, [2, "Bob", 25])
+await session.execute(insert_statement, [3, "Carol", 35])
+```
+
+Combine preparation with concurrent execution:
+
+```python
+insert_statement = await session.prepare(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)"
+)
+
+coroutines = [
+    session.execute(insert_statement, [i, f"user_{i}", i % 100])
+    for i in range(1000)
+]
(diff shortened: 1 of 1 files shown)


docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
 # Getting Started
 
-TBD.
\ No newline at end of file
+First make sure the driver has been properly installed, see [Installation](installation.md).
```
pasinskik: It should be: _First, make sure the driver has been properly installed. See [Installation](installation.md)._

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)",
+    [1, "Alice", 30]
+)
+```
+
+### Reading rows
+
+Rows are returned in pages as `dict[str, value]` by default. Iterate over all rows in all pages with transparent async paging with `async for`:
```
pasinskik: Sorry to nitpick! But I think it'll be better: _Iterate over all rows in all pages with transparent async paging using `async for`:_

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
 # Getting Started
 
-TBD.
\ No newline at end of file
+First make sure the driver has been properly installed, see [Installation](installation.md).
+
+> **Note:** This driver is **async-first** and requires Python's `asyncio`. All network operations - connecting, executing queries, fetching results - are coroutines that must be awaited inside an `async` function.
```
wprzytula:
```suggestion
> **Note:** This driver is **async-first** and requires Python's `asyncio`. All network operations - connecting, executing queries, fetching results - are coroutines, which must be awaited inside an `async` function.
```
This is a nondefining clause, as all coroutines must be awaited inside an `async` function in general.

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+## Connecting to a Cluster
+
+Before executing any queries you need to establish a connection to a Scylla cluster. This is done using `SessionBuilder`, which provides a chainable API for configuring the connection before calling `connect()`.
```
wprzytula: ScyllaDB

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+### Authentication
+
+Use `.user()` for plain-text authentication:
+
+```python
+session = await (
+    SessionBuilder()
+    .contact_points("127.0.0.2")
+    .user("cassandra", "cassandra")
```
wprzytula:
```suggestion
    .user("username", "password")
```

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+To set a default keyspace for the session, execute a `USE` statement after connecting:
+
+```python
+await session.execute(
+    "CREATE KEYSPACE IF NOT EXISTS my_ks "
+    "WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}"
+)
+await session.execute("USE my_ks")
+```
```
wprzytula, line 66: 💭 So there's no distinct `use_keyspace()` method exposed? This should be a low-hanging fruit, maybe let's implement it?

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+## Executing Queries
+
+Use `session.execute()` to run a CQL query. It is a coroutine - always `await` it:
+
+```python
+await session.execute(
+    "CREATE TABLE IF NOT EXISTS users (id int PRIMARY KEY, name text, age int)"
+)
+```
```
wprzytula: 🔧 Please avoid `query` naming - only `SELECT`s are queries. Use `statements` instead.

docs/source/topics/getting-started.md
wprzytula: Tip: use ` ```suggestion` blocks for better rendering of the code suggestions.

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+### Passing parameters
+
+Use `?` as the placeholder for bound parameters. Supply parameter values as a list or tuple:
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)",
+    [1, "Alice", 30]
+)
```
wprzytula, line 88: 🤔 Tuple has a lower overhead, right? Then let's encourage it over list in the example.

docs/source/topics/getting-started.md
wprzytula: 💭 The `value` put in the type annotation may confuse, since there's no such type AFAIK.

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+    for i in range(1000)
+]
+await asyncio.gather(*coroutines)
+```
+
+> Awaiting each `execute()` call sequentially in a loop is suboptimal - always prefer `asyncio.gather` or any other way of non-blocking execution when inserting or querying large amounts of data.
+
+## Prepared Statements
+
+Prepared statements are parsed by Scylla once and cached. Every subsequent execution only sends the bound parameter values, reducing network traffic and CPU usage.
```
wprzytula:
```suggestion
Prepared statements are parsed by Scylla once and cached. Every subsequent execution only sends the bound parameter values (and prepared statement id, which is a small hash instead of the full query string to yet be parsed), reducing network traffic and CPU usage.
```

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+statement = Statement("INSERT INTO users (id, name) VALUES (?, ?)").with_consistency(Consistency.Quorum)
+await session.execute(statement, [4, "Dave"])
+```
+
+To set consistency on a prepared statement:
+
+```python
+prepared = await session.prepare("INSERT INTO users (id, name) VALUES (?, ?)")
+prepared = prepared.with_consistency(Consistency.Quorum)
+```
```
wprzytula, line 167: ❓ 📌 
Does `prepare()` accept `Statement`, too?
- If no, do we have an issue for that?
- If yes:
    1. Can we show it, too?
    2. Will the `PreparedStatement` returned from `session.prepare(statement)` inherit settings (such as Consistency) that was set on `statement`?

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+### Request timeout
+
+Set a per-statement timeout in seconds:
+
+```python
+prepared = prepared.with_request_timeout(5.0)   # 5-second timeout
+```
```
wprzytula, line 175: 🔧 Let's briefly explain what's the semantics of this timeout. Let's be clear that after this timeout elapses, the execution fails immediately; and that this timeout spans all execution's retry attempts.

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+### Page size
+
+Control how many rows are fetched per page:
+
+```python
+statement = Statement("SELECT * FROM users").with_page_size(50)
+```
```
wprzytula, line 183: 💭  Let's maybe say what's the default page size, to give reader context.

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+**Session-wide default:**
+
+```python
+session = await (
+    SessionBuilder()
+    .contact_points("127.0.0.2")
+    .execution_profile(profile)
+    .connect()
+)
+```
```
wprzytula, line 216: ⛏️ Is this properly indented?

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+batch = Batch()
+batch.add(insert, [10, "Frank", 40])
+batch.add(insert, [11, "Grace", 28])
+batch.add(insert, [12, "Heidi", 33])
+
+# Optionally configure consistency on the batch itself
+batch = batch.with_consistency(Consistency.LocalQuorum)
+
+await session.batch(batch)
+```
```
wprzytula, line 319: 🔧 Missing section about execution profiles on batches.

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,387 @@
...
+
+## Configuring Logging
+
+The driver uses Python's standard `logging` module. Driver log messages are emitted under the `scylla` logger hierarchy.
```
wprzytula, line 323: > `scylla` logger hierarchy.

❓ What's that? Is this a usual Python naming?

wprzytula (changes requested):
Great job! 🚀

docs/source/topics/getting-started.md
pasinskik: > 💭 So there's no distinct `use_keyspace()` method exposed? This should be a low-hanging fruit, maybe let's implement it?

It has been implemented in [this pull request](https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/pull/67/changes/63f3f3b7f9df0713dc6c27fae6fe61b3ffc57af2), waiting to be merged.

docs/source/topics/getting-started.md
jakub-graczyk: image

for scylla logs logger uses `scylla` prefix

docs/source/topics/getting-started.md
jakub-graczyk: Yes, i. yes, ii. yes #103

docs/source/topics/getting-started.md
jakub-graczyk: Oh, I also created one oops!

wprzytula: As @Lorak-mmk noted correctly, those are different methods. One sets the keyspace to be used since the beginning of Session's lifetime, the other sets it dynamically while the Session is alive.

docs/source/topics/getting-started.md
```diff
@@ -1,3 +1,384 @@
...
+## Working with Keyspaces
+
+To set a default keyspace for the session a `use_keyspace` method is available:
+
+```python
+await session.execute(
+    "CREATE KEYSPACE IF NOT EXISTS my_ks "
+    "WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}"
+)
+await session.use_keyspace("my_ks")
```
wprzytula, line 65: 🤔 Doesn't this method accept two arguments, with second being boolean (`case_sensitive`)?

wprzytula (approved):

docs/source/topics/getting-started.md
jakub-graczyk: It does but it's default is false so it doesn't need to be specified.

PR #102, opened 2026-05-29
add: use keyspace method on session

jakub-graczyk:
Adds the method, tests and renames manual `session.execute("USE keyspace_name") to use the new method.


src/session.rs
```diff
@@ -44,6 +44,15 @@ impl TryFrom<Arc<Session>> for PySession {
...
 impl PySession {
+    async fn use_keyspace(&self, keyspace: &str) -> Result<RequestResult, DriverExecuteError> {
+        self.execute_unpaged(
+            ExecutableStatement::Unprepared(Statement::new(format!("USE {};", keyspace))),
+            PyValueList::Empty,
+            None,
+        )
+        .await
+    }
+
```
Lorak-mmk, line 60: This is wrong. In Rust Driver we have a `use_keyspace` method on a `Session`, you should use that.
Also, you should add `case_sensitive` parameter.

Lorak-mmk (changes requested):

src/session.rs
jakub-graczyk: What are the advantages of that method instead of using a "USE *" statement?

wprzytula: It validates the keyspace, preventing CQL injections, among others.

Lorak-mmk: When you issue a `USE <ks>` statement, driver won't immediately know that it is a `USE` statement (because driver doesn't parse statements).
It will first send it as a normal request, receive a response indicating it was a USE statement, and only then proceed to its internal handling of USE (sending it on all connections).
With `use_keyspace` you eliminate this first hop, driver can proceed switching all connections immediately.

Lorak-mmk (approved):

python/scylla/_rust/session.pyi
```diff
@@ -17,6 +17,15 @@ class Session:
...
         """
         ...
+    async def use_keyspace(self, keyspace: str) -> RequestResult:
+        """
+        Sends `USE <keyspace>` request on all connections
+        This allows to write `SELECT * FROM table` instead of `SELECT * FROM keyspace.table`
+
+        Note that even failed `use_keyspace` can change currently used keyspace - the request is sent on all connections and
+        can overwrite previously used keyspace.
+        """
```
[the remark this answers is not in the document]
wprzytula: This is incorrect.

python/scylla/errors.py
```diff
@@ -43,4 +46,6 @@
     "ValueOverflowSerializationError",
     "SerializeFailedError",
     "PySerializationFailedError",
+    "UseKeyspaceError",
+    "ClusterStateTokenError",
```
wprzytula, line 55: ❓ Why is `ClusterStateTokenError` added here in this commit?

pasinskik: Because it was accidentally omitted in the metadata PR.

src/errors.rs
```diff
@@ -1173,6 +1175,64 @@ impl From<DriverSerializationError> for scylla::serialize::SerializationError {
     }
 }
```
pasinskik, line 1182: I think it's great practice to have a doc comment for the enum, as well as for the variants (like all the other driver error types).

Something like this:


```suggestion
/// Errors that can occur during use_keyspace operation on a session object.
```

src/errors.rs
```diff
@@ -1173,6 +1175,64 @@ impl From<DriverSerializationError> for scylla::serialize::SerializationError {
...
+#[derive(Debug)]
+pub(crate) enum DriverUseKeyspaceError {
+    BadKeyspaceName { message: String },
+    RequestError { message: String },
+    KeyspaceNameMismatch { message: String },
+    RequestTimeout { message: String },
+    RuntimeTaskJoinFailed { message: String },
+}
+
+impl From<RustUseKeyspaceError> for DriverUseKeyspaceError {
```
pasinskik, line 1193: I think this mapping is too loosy. If we map 1-1 from the Rust Driver's enum - we should preserve the relevant information.

i.e. `KeyspaceNameMismatch` and `RequestTimeout` carry additional attributes that are collapsed here into `message: String`.

The current enum variants are no better than just storing the original `RustUseKeyspaceError`, because they do not distinguish data and all are eventually mapped to the same Python exception anyway. Either we should preserve the richer context or possibly directly wrap `RustUseKeyspaceError`.

Moreover, Rust driver also uses a nested enum to represent `BadKeyspaceName`. We could preserve this structure or collapse it if you deem this irrelevant.

pasinskik (changes requested):

python/scylla/errors.py
wprzytula: In such case, it should be added in a separate commit, with proper description.

python/tests/test_use_keyspace.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+    await session.execute("CREATE TABLE IF NOT EXISTS test_keyspace.test_table (id int PRIMARY KEY)")
+
+    await session.use_keyspace("test_keyspace")
+    await session.execute("INSERT INTO test_table (id) VALUES (1)")
+
+    await session.use_keyspace("other_keyspace")
+
+    with pytest.raises(ExecuteError):
+        await session.execute("INSERT INTO test_table (id) VALUES (1)")
```
wprzytula: :pushpin:  We're using hardcoded keyspace names, and we're not dropping them at the end of each test. This may have the following consequences:
1. Test runs are not idempotent.
2. Tests may interfere with each other when (unintentionally) using same keyspace.

@Lorak-mmk How do we want to approach this?
In Rust Driver, we have both solutions:
1. Tests drop keyspaces they create, so cleanup is ensured.
2. Tests use unique keyspace names, generated from timestamps.

Lorak-mmk: And we should do both of those things here as well.

Btw I thought other tests already have some helper utils to drop created keyspaces? Did I imagine that?

PR #104, opened 2026-06-01
Add: Documentation for query results

patrycja-ziemkiewicz:
Fixes: https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/89

## Pre-review checklist

- [x] I have adjusted the documentation in `./docs/source/`.

### DIFF
--- a/docs/source/topics/using/query-results.md
+++ b/docs/source/topics/using/query-results.md
@@ -1,0 +1,306 @@
+# Query results
+
+## Overview
+
+`Session.execute()` returns a `RequestResult`.
+
+For `SELECT` statements, the result contains rows. By default, paging is enabled, so rows are fetched page by page instead of loading the whole result set in one response.
+
+For statements that do not return rows, such as `INSERT`, `UPDATE`, `DELETE`, `CREATE TABLE`, or `DROP TABLE`, the returned `RequestResult` can usually be ignored after the request succeeds.
+
+## RequestResult
+
+### Row representation
+
+By default, each row is represented as a dictionary:
+
+```python
+{
+    "a": 1,
+    "b": 2,
+    "c": "abc",
+}
+```
+
+The exact row representation can be changed with a custom `RowFactory`. See [Custom row shapes with RowFactory](#custom-row-shapes-with-rowfactory).
+
+### NULL values
+
+CQL `null` values are represented as `None` in Python.
+
+```python
+result = await session.execute("SELECT a, b, c FROM select_paging")
+
+async for row in result:
+    value = row["c"]
+
+    if value is None:
+        print("column c is NULL")
+    else:
+        print(value)
+```
+
+### Immutability
+
+`RequestResult` represents one immutable query result state.
+
+Methods that need to move forward through the result set do not mutate the existing object. For example, `fetch_next_page()` returns a new `RequestResult` containing the next page.
+
+```python
+result = await session.execute(stmt)
+
+while True:
+    for row in result.iter_current_page():
+        print(row)
+
+    next_result = await result.fetch_next_page()
+    if next_result is None:
+        break
+
+    result = next_result
+```
+
+This is important because the original `RequestResult` still represents the page it was created with. If you manually fetch another page, assign the returned value to a new variable or replace the old one.
+
+## Paging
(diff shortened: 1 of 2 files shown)


docs/source/topics/using/query-results.md
```diff
@@ -0,0 +1,286 @@
+# Query results
+
+`Session.execute()` returns a `RequestResult`.
+
+For `SELECT` statements, the result contains rows. By default, paging is enabled, so rows are fetched page by page instead of loading the whole result set in one response.
+
+For statements that do not return rows, such as `INSERT`, `UPDATE`, `DELETE`, `CREATE TABLE`, or `DROP TABLE`, the returned `RequestResult` can usually be ignored after the request succeeds.
```
wprzytula, line 9:
```suggestion
For statements that do not return rows, such as `INSERT`, `UPDATE`, `DELETE`, `CREATE TABLE`, or `DROP TABLE`, the returned `RequestResult` still contain some information, including warnings returned by the DB and the tracing id if tracing was enabled.
```

This is so on the Rust side.
But this probably isn't yet implemented?

docs/source/topics/using/query-results.md
```diff
@@ -0,0 +1,286 @@
...
+Fetching a large result set without paging can put unnecessary pressure on both the driver and the database. If the result is large, or if the query needs to scan many tombstones, an unpaged query may:
+
+- increase latency,
+- use much more memory in the driver,
+- put higher load on the cluster,
+- be more likely to time out.
+
+To sum up, for SELECTs (especially those that may return a lot of data) prefer paged queries.
+
+If you explicitly want to execute an unpaged query, pass paged=False:
```
wprzytula:
```suggestion
If you explicitly want to execute an unpaged query, pass `paged=False`:
```

docs/source/topics/using/query-results.md
patrycja-ziemkiewicz: Currently request result does not contain any information except from the returned rows.

docs/source/topics/using/query-results.md
```diff
@@ -0,0 +1,306 @@
...
+
+By default, each row is represented as a dictionary:
+
+```python
+{
+    "a": 1,
+    "b": 2,
+    "c": "abc",
+}
+```
```
wprzytula, line 23: 🔧 Let's explicitly say that keys are columns' names.

docs/source/topics/using/query-results.md
```diff
@@ -0,0 +1,306 @@
...
+Paging is enabled by default and should usually stay enabled for `SELECT` queries.
+
+Fetching a large result set without paging can put unnecessary pressure on both the driver and the database. If the result is large, or if the query needs to scan many tombstones, an unpaged query may:
+
+- increase latency,
+- use much more memory in the driver,
+- put higher load on the cluster,
+- be more likely to time out.
+
+To sum up, for SELECTs (especially those that may return a lot of data) prefer paged queries.
```
wprzytula: ⛏️ I really liked the pinch of humour I put in the Rust Driver docs - it reaches readers' awareness easier IMO. Can we preserve it?

wprzytula (approved):

docs/source/topics/using/query-results.md
wprzytula: 💭 Conditional UPDATEs (batch LWT) return a single pseudo-row with boolean `applied` field, saying whether the statement was applied or not. Then one might like to examine the returned result of an INSERT statement.
https://docs.scylladb.com/manual/stable/features/lwt.html#what-row-does-if-clause-apply-to

docs/source/topics/using/query-results.md
```diff
@@ -0,0 +1,306 @@
...
+```python
+result = await session.execute("SELECT a, b, c FROM select_paging")
+
+async for row in result:
+    value = row["c"]
+
+    if value is None:
+        print("column c is NULL")
+    else:
+        print(value)
```
wprzytula, line 40:
```suggestion
    if value is None:
        print("column c is NULL")
    else:
        print("column c has value", value)
```

PR #105, opened 2026-06-02
Add retry policy

paulinaczajkowska:
## Pre-review checklist


src/lib.rs
```diff
@@ -69,5 +70,6 @@ fn scylla(py: Python<'_>, module: &Bound<'_, PyModule>) -> PyResult<()> {
     add_submodule(py, module, "other_policies", other_policies::other_policies)?;
     add_submodule(py, module, "cluster", cluster::cluster)?;
     add_submodule(py, module, "routing", routing::routing)?;
+    add_submodule(py, module, "retry_policy", policies::retry::retry_policy)?;
```
jakub-graczyk, line 73: Take a look at cluster and it's metadata submodule creation. Let's register nested submodules as submodules in their parent, not on the top level.

python/scylla/_rust/retry_policy.pyi
```diff
@@ -0,0 +1,213 @@
...
+    is_idempotent: bool
+    consistency: Consistency
+
+    def __init__(self, error: RequestAttemptError, is_idempotent: bool, consistency: Consistency) -> None: ...
+
+@runtime_checkable
+class CustomRetrySession(Protocol):
+    def __init__(self) -> None: ...
+    def decide_should_retry(self, request_info: RequestInfo) -> RetryDecision: ...
+    def reset(self) -> None: ...
```
patrycja-ziemkiewicz, line 507: It would be better to name protocol RetrySession as all other builtin RetrySession policies also implements this protocol

python/scylla/_rust/retry_policy.pyi
```diff
@@ -0,0 +1,213 @@
...
+
+class FallthroughRetrySession:
+    def __init__(self) -> None: ...
+    def decide_should_retry(self, request_info: RequestInfo) -> RetryDecision: ...
+    def reset(self) -> None: ...
+
+@runtime_checkable
+class CustomRetryPolicy(Protocol):
+    def __init__(self) -> None: ...
+    def new_session(self) -> RetrySession: ...
```
patrycja-ziemkiewicz, line 536: Same here, It would be better to name it RetryPolicy

python/scylla/_rust/retry_policy.pyi
```diff
@@ -0,0 +1,213 @@
...
+
+RetryPolicy: TypeAlias = (
+    DefaultRetryPolicy | DowngradingConsistencyRetryPolicy | FallthroughRetryPolicy | CustomRetryPolicy
+)
+
+RetrySession: TypeAlias = (
+    DefaultRetrySession | DowngradingConsistencyRetrySession | FallthroughRetrySession | CustomRetrySession
+)
```
patrycja-ziemkiewicz: There is no need for creating type aliases as all of these builtin policies implement either CustomRetryPolicy or CustromRetrySession protocol

src/policies/retry/request.rs
```diff
@@ -0,0 +1,55 @@
...
+        consistency: PyConsistency,
+    ) -> Self {
+        Self {
+            error: error.clone(),
+            is_idempotent,
+            consistency,
+            rust_error: error.into(),
+        }
+    }
+}
```
patrycja-ziemkiewicz, line 55: In Rust Driver it is not possible for users to create ```RequestInfo``` do we want to enable it in Python-RS driver?

src/policies/retry/policies.rs
```diff
@@ -0,0 +1,252 @@
...
+        let mut inner = self._inner.lock().unwrap();
+        inner
+            .decide_should_retry(request_info.to_request_info())
+            .into()
+    }
+
+    fn reset(&self) {
+        let mut inner = self._inner.lock().unwrap();
+        inner.reset();
+    }
```
patrycja-ziemkiewicz: Use ```lock_py_attached()``` method on mutex to avoid deadlocks with Python interpreter.

src/policies/retry/policies.rs
```diff
@@ -0,0 +1,252 @@
...
+        Python::attach(|py| {
+            let py_retry_session = self._inner.bind(py);
+            let py_request_info = PyRequestInfo::from(&request_info);
+
+            let result = py_retry_session
+                .call_method1(intern!(py, "decide_should_retry"), (py_request_info,));
+
+            match result {
+                Ok(res) => match res.extract::<PyRetryDecision>() {
+                    Ok(py_retry_decision) => RetryDecision::from(py_retry_decision),
```
patrycja-ziemkiewicz: Please use `cast` instead of `extract` when you only need a typed reference to a Python object.

src/policies/retry/policies.rs
```diff
@@ -0,0 +1,252 @@
...
+                }
+            }
+        })
+    }
+
+    fn reset(&mut self) {
+        Python::attach(|py| {
+            let obj = self._inner.bind(py);
+
+            if let Err(err) = obj.call_method0("reset") {
```
patrycja-ziemkiewicz: You could use ```intern!()``` macro here.

src/policies/retry/policies.rs
```diff
@@ -0,0 +1,252 @@
...
+
+impl RetryPolicy for PyCustomRetryPolicy {
+    fn new_session(&self) -> Box<dyn RetrySession> {
+        Python::attach(|py| -> Box<dyn RetrySession> {
+            let policy = self._inner.bind(py);
+
+            match policy.call_method0(intern!(py, "new_session")) {
+                Ok(session) => Box::new(PyCustomRetrySession {
+                    _inner: session.unbind(),
+                }),
```
patrycja-ziemkiewicz, line 157: Here method ```new_session``` might return builtin implementation of RetryPolicy. In this case We could avoid calling python functions for the classes implemented in Rust by first trying to cast it to some of the builtins policies and if We do not succeed We can assume user provided a custom retry session.

src/policies/retry/policies.rs
```diff
@@ -0,0 +1,252 @@
...
+    } else if obj.is_instance_of::<PyDowngradingConsistencyRetryPolicy>() {
+        Arc::new(DowngradingConsistencyRetryPolicy::new())
+    } else if obj.is_instance_of::<PyFallthroughRetryPolicy>() {
+        Arc::new(FallthroughRetryPolicy::new())
+    } else {
+        Arc::new(PyCustomRetryPolicy {
+            _inner: rp.clone_ref(py),
+        })
+    }
+}
```
patrycja-ziemkiewicz, line 250: Instead of creating a new Rust retry policy for built-in Python policy wrappers, could we reuse the existing inner policy by cloning its `Arc`?

python/scylla/_rust/retry_policy.pyi
patrycja-ziemkiewicz, line 1: Please add docstrings

src/policies/retry/request.rs
wprzytula: What for? @paulinaczajkowska

paulinaczajkowska: Right now I'm using `RequestInfo` in tests for the retry policy, but maybe I can write them in Rust (like cache tests) and it won't be necessary to expose `new()` as a Python method.

PR #106, opened 2026-06-05
fix: column type inheritance type hints

jakub-graczyk:
Correctly shows inheritance structure in type hints.


Lorak-mmk (approved):
Thanks!

PR #109, opened 2026-06-07
Add statement values documentation

patrycja-ziemkiewicz:
Fixes: https://github.com/scylladb-zpp-2025-python-rs-driver/python-rs-driver/issues/88

## Pre-review checklist

### DIFF
--- a/docs/source/topics/using/statement-values.md
+++ b/docs/source/topics/using/statement-values.md
@@ -1,0 +1,99 @@
+# Statement values
+
+Statement text is constant, while values may change between executions. Values can be passed separately from the statement text using bound parameters.
+
+Values can be provided as:
+
+- a list,
+- a tuple,
+- a mapping,
+- or omitted when the statement does not need any values.
+
+Bound parameters should be used instead of building queries by concatenating strings. This keeps the query easier to read and avoids problems such as invalid escaping or injection vulnerabilities.
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)",
+    (user_id, "Alice", 30),
+)
+```
+
+## Passing positional values
+
+Use positional values when the statement contains positional bind markers, written as `?`.
+
+Each `?` in the statement is filled with the matching value from the provided sequence, in order. The sequence may be a tuple:
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)",
+    (user_id, "Alice", 30),
+)
+```
+
+or a list:
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)",
+    [user_id, "Alice", 30],
+)
+```
+
+Values may also be passed as a mapping, such as a dictionary. When a mapping is used, dictionary keys are matched by name. The order of keys does not matter:
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (:id, :name, :age)",
+    {
+        "id": user_id,
+        "name": "Alice",
+        "age": 30,
+    },
+)
+
+await session.execute(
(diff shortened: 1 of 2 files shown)


docs/source/topics/using/statement-values.md
```diff
@@ -0,0 +1,99 @@
...
+- or omitted when the statement does not need any values.
+
+Bound parameters should be used instead of building queries by concatenating strings. This keeps the query easier to read and avoids problems such as invalid escaping or injection vulnerabilities.
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (?, ?, ?)",
+    (user_id, "Alice", 30),
+)
+```
```
wprzytula: _...as well as improves routing and performance of requests._

docs/source/topics/using/statement-values.md
```diff
@@ -0,0 +1,99 @@
...
+
+```python
+await session.execute(
+    "INSERT INTO users (id, name, age) VALUES (:id, :name, :age)",
+    {
+        "id": user_id,
+        "name": "Alice",
+        "age": 30,
+    },
+)
```
wprzytula, line 53: 🤔 It's not clear what are the relationships between each of three names:
1. `id` in column name specification;
2. `id` in value name specification;
3. `id` as a key in the dictionary.

Must 2 and 3 match? What about 1 and 2?

docs/source/topics/using/statement-values.md
```diff
@@ -0,0 +1,99 @@
...
+
+If the statement does not contain any bind markers, values can be omitted:
+
+```python
+await session.execute("SELECT * FROM users")
+```
+
+## Prepared statements and statement objects
+
+Prepared statements and statement objects accept values in the same way as plain query strings.
```
wprzytula: 🤔 Unjustified discrepancy. _Prepared statements_ vs _statement objects_. Why not _Prepared and unprepared statement objects_ or _Prepared and unprepared statements_?

docs/source/topics/using/statement-values.md
```diff
@@ -0,0 +1,99 @@
...
+    prepared,
+    (user_id, "Alice", 30),
+)
+```
+
+## Type conversion
+
+Python values are converted to the CQL types expected by the database schema.
+
+For the full mapping of supported CQL and Python types, see [Data Types](../data-types.md).
```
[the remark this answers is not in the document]
wprzytula: Why is this marked as resolved?

wprzytula (changes requested):

docs/source/topics/using/statement-values.md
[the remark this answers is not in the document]
patrycja-ziemkiewicz: Because this link will became valid when we merge data-types PR.

docs/source/topics/using/statement-values.md
patrycja-ziemkiewicz: The dictionary keys must match the bind marker names, without the leading :. So :id is matched by the "id" key.

The bind marker names do not have to match the column names. They only do here because it makes the example shorter. This would also be valid:

```
await session.execute(
    "INSERT INTO users (id, name, age) VALUES (:user_id, :user_name, :user_age)",
    {
        "user_id": user_id,
        "user_name": "Alice",
        "user_age": 30,
    },
)
```

I can update the example to use different bind marker names, so this relationship is clearer.

Lorak-mmk: > 
> ```
> await session.execute(
>     "INSERT INTO users (id, name, age) VALUES (:user_id, :user_name, :user_age)",
>     {
>         "user_id": user_id,
>         "user_name": "Alice",
>         "user_age": 30,
>     },
> )
> ```
> 
> I can update the example to use different bind marker names, so this relationship is clearer.

This example is imo better

